PR #7, opened 2021-02-14
Update to latest gotrue-py, monkey patch for sync behaviour, add working tests

fedden:

Adds:
* Sync `.execute()` monkey patch for easier calling of supabase.select/etc
* Test for `select()` client code to demonstrate it working
* Updates the client to work with the new gotrue-py changes (needs to be merged first).
* Adds two hotfixes for realtime module
* Enables sync insertion of row via `insert()` monkey patch.
* Adds test for `insert()` client code to demonstrate it working.

Current behaviour is that the select op doesn't work as is and on top of that requires async code which in the authors opinion is not a requirement we should enforce on users. 

The client also wont correctly call the gotrue-py library unless these changes are merged.


fedden:
Screenshot 2021-02-14 at 21 01 08

Local tests are passing. Currently CI is busted for master, develop and this feature branch. We should get to that when we can

lqmanh:
@fedden, it seems to have something wrong with `supabase-realtime-py` version. Could you have a look at it?

fedden:
@lqmanh thanks for checking this PR out. As far as I can see, according to the diff this PR doesn't change [the version of realtime-py](https://github.com/supabase/supabase-py/blob/2a9bfaf3b8fc19d87590a8fb4c5de5b637ec03c5/pyproject.toml#L11). I would find it very helpful if you could kindly ellaborate on the issue you've encountered!

cc @J0 who has been working directly on realtime-py

Thanks!

fedden:
Oh I see what you mean, it's the failing tests

fedden:
Probably just needs a deploy @J0 ? Regardless I don't think it was this PR that has introduced the fault as the version is unchanged.

fedden:
I have had a few spare minutes this sundaym so have added support for `insert()`, added a test in test_client.py to demonstrate this, and have ran tests locally to confirm it is working. 

Screenshot 2021-03-28 at 12 29 37


To be completely honest I don't have the spare time to look into realtime-py and it's also not something I have been working on nor familiar with. I'm pretty sure until that is deployed however, the tests for this (and any other branch) will fail.

J0:
> Probably just needs a deploy @J0 ? Regardless I don't think it was this PR that has introduced the fault as the version is unchanged.

Apologies, just saw this comment. Will take a look in a bit. Thanks for catching this!

pvbhanuteja:
Any updates it still seems to fail @J0

J0:
Hey @pvbhanuteja,

Sorry for the delay and thank you for your continued patience. Was a little preoccupied with work. I'll look at it in two and a half hours and push a patch by tonight.

Joel

J0:
For reference, `supabase-realtime-py` was the alternate namespace we used until we managed to get hold of the `realtime-py` namespace. I have updated the `pyproject.toml` to reflect this.

fedden:
Cool I've deployed the latest go-true and have set the tests to use my supabase isntance. I can't add these env vars as I don't have admin access to github workflows

fedden:
But importantly, the tests I've introduced are passing locally and remotely. We can iterate on the specifics and improve this, but it's an improvement on what we had before and we can add to them for each new feature branch

PR #25, opened 2021-05-19
Reduce test code duplication via supabase Client in pytest fixture

olirice:

The test suite has some minor setup duplication at the top of each function. Namely,

```python
url: str = os.environ.get("SUPABASE_TEST_URL")
key: str = os.environ.get("SUPABASE_TEST_KEY")
supabase: Client = create_client(url, key)
```

This change factors that test setup into a pytest fixture that is created once and shared across the tests that require a `supabase.Client`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,7 @@ import pytest
 from supabase_py import Client, create_client
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="session")
 def supabase() -> Client:
     url: str = os.environ.get("SUPABASE_TEST_URL")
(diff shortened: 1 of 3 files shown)


J0:
Hey @olirice ,

Thank you for your contribution! This is helpful and will save us time looking through the duplicated sections of test code. Can we check if there are further changes that you'd like to make? Think we can review and  subsequently merge otherwise

olirice:
Hi @J0 

This was the only immediate change I had on the test suite
There is also #26 (if there is interest in that change) but that belong in a separate PR

J0:
Cool, thanks!

PR #28, opened 2021-05-22
Add pre-commit hooks enforcing a standard style

### LINKED ISSUE #26: Enforce code consistency via pre-commit
# Feature request

Is there any interest in hooking up [pre-commit](https://pre-commit.com/) with the usual battery hooks for python libraries e.g. [black](https://github.com/psf/black), [isort](https://pycqa.github.io/isort/), [autoflake](https://github.com/myint/autoflake), and [mypy](https://github.com/python/mypy), to enforce code consistency/standards?

If so, I'm happy to open a PR

olirice:

Enforces code consistency via pre-commit hooks and checks pushed code on all branches in CI

I left off the mypy check for now because there were some tricky type errors in the query builder 

Resolves #26

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,27 @@
+repos:
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+    - id: trailing-whitespace
+    - id: check-added-large-files
+    - id: mixed-line-ending
+      args: ['--fix=lf']
+
+-   repo: https://github.com/pre-commit/mirrors-isort
+    rev: v5.8.0
+    hooks:
+    -   id: isort
+        args: ['--multi-line=3', '--trailing-comma', '--force-grid-wrap=0', '--use-parentheses', '--line-width=88']
+
+-   repo: https://github.com/humitos/mirrors-autoflake.git
+    rev: v1.1
+    hooks:
+    -   id: autoflake
+        args: ['--in-place', '--remove-all-unused-imports']
+
+-   repo: https://github.com/ambv/black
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - id: trailing-whitespace
+    - id: check-added-large-files
+    - id: mixed-line-ending
+      args: ['--fix=lf']
+
+-   repo: https://github.com/pre-commit/mirrors-isort
+    rev: v5.8.0
+    hooks:
+    -   id: isort
+        args: ['--multi-line=3', '--trailing-comma', '--force-grid-wrap=0', '--use-parentheses', '--line-width=88']
```
olirice, line 15: Calling this out because the settings look strange at first.

These are the recommended settings for interop with black (which also may adjust imports)

olirice:
The black hook is going to cause conflicts any time other stuff merges into develop before this PR. I can strip out the commit that applies the changes and limit this PR to just the pre-commit config, CI, and README files if you prefer.

The CI hooks would fail after merge but you'd be able to apply the hooks locally when you're ready. Would that be easier to deal with?

J0:
Hey @olirice,

Thanks for considering the downstream consequences of this PR, it is really thoughtful of you. No worries, I'll look into this Pull Request in a few hours so that we can get this merged before anything else gets merged.

Joel

.pre-commit-config.yaml
J0: Makes sense. Thanks for pointing this out! Just going to plop [the reference doc](https://pycqa.github.io/isort/docs/configuration/profiles/) here for my own reference.

J0:
LGTM

Issue #30, opened 2021-05-31
'Error in $: not enough input' on .update()


rphly:
# Bug report

## To Reproduce
```
res = db.table(table_name).update({"notes": notes}).eq("id", user_id).execute()
```
## Expected behavior

Update and return me the row.

## Actual behaviour

`{'data': {'message': 'Error in $: not enough input'}, 'status_code': 400}`

## System information

- Version of supabase-py: 0.0.2

rphly:
Temporary workaround is to use .insert() with upsert=True

fedden:
Yes noticing this too. Is upsert actually working or is it simply inserting a new row? From what I've noticed it's inserting new rows so would be helpful to get confirmation thats what you are seeing. Perhaps request headers/jwt is misconfigured somehow..

rphly:
@fedden it updates the row when I use insert with upsert=true.

remybroun:
It would be nice to have this documented somewhere. Without this issue there is no way to know how to update a row.

J0:
Hey @remybroun @rphly,

We have tested this issue and it seems to work fine now.  We are closing this for now but please feel free to let us know if you're still facing any issues. 
![image](https://user-images.githubusercontent.com/8011761/139558552-488da81c-743a-4664-b9ab-b5dd485b0ec7.png)

Thanks!

Krish-Nerkar:
> Hey @remybroun @rphly,
> 
> We have tested this issue and it seems to work fine now. We are closing this for now but please feel free to let us know if you're still facing any issues. ![image](https://user-images.githubusercontent.com/8011761/139558552-488da81c-743a-4664-b9ab-b5dd485b0ec7.png)
> 
> Thanks!

I am still facing the same issue :(

`res = supabase.table("Servers").update({
                    "Server_ID" : previous_data['data'][0]['Server_ID'],
                    "Server_Name" : previous_data['data'][0]['Server_Name'],
                    "Channel_ID" : previous_data['data'][0]['Channel_ID'],
                    "Categories" : previous_data['data'][0]['Categories'], 
                    "Links": previous_links
                }).execute()`

![image](https://user-images.githubusercontent.com/49949733/139632439-97d83268-c5df-41f7-b587-8444616a9790.png)

dreinon:
Make sure to update your supabase version to the latest and tell us if you still face the issue please 😊

mohnish7:
[...]

PR #41, opened 2021-08-18
fix: missing json bodies in patch and put requests

darora:
Fixes https://github.com/supabase/postgrest-py/issues/7 and #30 

Why are we monkey patching postgrest-py here? If we really need to hide the async reality and offer a synchronous interface, can we move it down to the postgrest-py layer so at least it's tested and kept in sync?

### DIFF
--- a/supabase_py/lib/query_builder.py
+++ b/supabase_py/lib/query_builder.py
@@ -18,8 +18,10 @@ def _execute_monkey_patch(self) -> Dict[str, Any]:
         additional_kwargs = {"json": self.json}
     elif method == "put":
         func = requests.put
+        additional_kwargs = {"json": self.json}
     elif method == "patch":
         func = requests.patch
+        additional_kwargs = {"json": self.json}
     elif method == "delete":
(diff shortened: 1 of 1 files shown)


fedden:
Agreed on moving the requests/sync stuff to the `postgrest-py` lib. 

> Why are we monkey patching postgrest-py here?

It was supposed to be a temporary measure to provide an easy interface for the supabase users so we could get a release quickly out of the door, but agreed moving this to postgrest would be preferable.

supabase_py/lib/query_builder.py
```diff
@@ -18,8 +18,10 @@ def _execute_monkey_patch(self) -> Dict[str, Any]:
         additional_kwargs = {"json": self.json}
     elif method == "put":
         func = requests.put
+        additional_kwargs = {"json": self.json}
     elif method == "patch":
         func = requests.patch
+        additional_kwargs = {"json": self.json}
     elif method == "delete":
```
fedden, line 25: lgtm thanks!

Issue #42, opened 2021-08-23
X-Client-Info header


awalias:
we're rolling out a header similar to Stripe's App-Data in all of the client libs so issues can be debugged more easily between the client and the backend

the javascript libs are done already: https://github.com/supabase/supabase-js/pull/238

the format is: `X-Client-Info: supabase-js/1.11.0`

for client libs that wrap others (like how supabase-js wraps gotrue-js) we allow the wrapper lib to **overwrite** the wrapped lib's header (given that we can infer the gotrue-js version based on which supabase-js was used to make the call)

any help with rolling this out here would be incredible

dreinon:
May this issue be closed already?

PR #43, opened 2021-08-26
Add some default headers to wrapped client libs

lqmanh:

Add some default headers to wrapped client libs.

See #42.

## Additional context

Some other minor changes included:
- Add Black as a dev dependency, because we use it in the pre-commit hook.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ license = "MIT"
 
 [tool.poetry.dependencies]
 python = "^3.7.1"
-postgrest-py = { git = "https://github.com/supabase/postgrest-py.git" }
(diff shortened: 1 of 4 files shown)


lqmanh:
@fedden Would you mind having a quick look at this?

J0:
LGTM but will await @fedden 's review

J0:
@lqmanh think @fedden might be busy. If you'd be okay with it let's wait a day and merge?

Let me know what you think!
- JL

lqmanh:
@J0 Uhm I believe that's fine. It's pretty similar to `supabase-js` version.

J0:
merged, Thanks!

PR #53, opened 2021-10-01
Type hinting on a single file.

SwarajBaral:
hey @J0. I tried adding a bit of type hinting to the file `supabase_py/lib/query_builder.py`. Please forgive me if I made any foolish errors. Do let me know if I should continue. And it will mean a lot if you could list down some resources as well.

### DIFF
--- a/supabase_py/lib/query_builder.py
+++ b/supabase_py/lib/query_builder.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any, Callable, Dict
 
 import requests
 from httpx import AsyncClient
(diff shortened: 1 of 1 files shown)


dreinon:
I find it pretty well done for now.

Remember that standard python types like dict, list, etc, admit type indexing since a few python versions ago. We might want to use those instead of the ones from the `typing` library.

Also, for more complex structures like headers (and in order to make the code more readable and documented), we might want to use something like `pydantic` or `dataclasses` to specify the exact structure. Or, if that's too much, for this example, we can define a type `HeadersType = dict[str, str]` and use that as the type.

What do you think @J0 ?

SwarajBaral:
Thank you so much for the detailed explanation @dreinon. And I will look into how to define the complex data structured objects. I will keep you posted.

dreinon:
@SwarajBaral how are you doing with this task? Do you need any kind of help?

Thanks!

dreinon:
Don't get me wrong @SwarajBaral! I didn't mean to hurry you up, so no worries! Just making sure you were fine and you didn't need any help :)

dreinon:
Hey @SwarajBaral! We are closing this due to inactivity, but feel free to reopen it or submit another PR with more changes. Maybe you can talk to @leynier and @anand2312 if you still want to work on this and you guys can split the work 🙂
Thanks for your contribution though!

PR #55, opened 2021-10-04
chore: supabase_py -> supabase

J0:


supabase/__init__.py
```diff
@@ -0,0 +1,6 @@
+__version__ = "0.0.2"
```
dreinon, line 1: Shouldn't this be 0.0.3??

J0: Yup, good catch. It should be 0.0.3 Thanks!

dreinon:
Why isn't this PR being merged already?

J0:
@dreinon, just haven't been active. Merging now, thanks! Will group my issues together so I can respond faster.

PR #61, opened 2021-10-13
Async storage buckets

anand2312:
This PR adds async support  to the `storage buckets API` by switching to `httpx`. All sync methods can now also be called with async/await syntax, by passing `is_async=True` to the `StorageBucketAPI` constructor (this value defaults to False).

Other smaller changes:
- Removed version pin from pre-commit-config (would error out if you did not have a local install of Python 3.7)
- Added `name`, `public` arguments to `StorageBucketAPI.create_bucket` method
- Better error propagation by raising a `StorageException`, instead of excepting all errors and printing them out.
- Type-hinted the `storage_bucket_api.py` file (I might've gone a little overboard with that, sorry)

### DIFF
--- a/supabase/lib/storage/storage_bucket_api.py
+++ b/supabase/lib/storage/storage_bucket_api.py
@@ -91,11 +91,9 @@ class StorageBucketAPI:
         response_data = response.json()
 
         if not response_class:
-            # if no response_class is specified, return the raw response
             return response_data
 
         if isinstance(response_data, list):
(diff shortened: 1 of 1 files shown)


anand2312:
I agree, it's not the neatest solution out there, I'm open to suggestions. I was mainly trying to reduce the amount of code duplication that would occur with two separate classes.
I can split it into two classes if more people are leaning towards that direction

Issue #65, opened 2021-10-15
Cannot create a bucket using the storage client


julianolf:
**Describe the bug**
The method from the storage client responsible for creating new buckets is not working properly, it just prints a generic HTTP error message.

**To Reproduce**
1. Create a client instance.
2. Retrieve the storage client.
3. Execute the `create_bucket` method.
```python
from supabase import create_client

url = "https://xxx.supabase.co"
key = "xxx"
client = create_client(url, key)
storage = client.storage()
storage.create_bucket("mybucket")
```

**Expected behavior**
The bucket is not created and the following message is printed out.
```
HTTP error occurred: 400 Client Error: Bad Request for url: https://xxx.supabase.co/storage/v1/bucket
```

**Additional context**
After debugging the request performed by the method I could see a more detailed error message.

The Supabase API was returning the following payload:
```json
{
  "statusCode": "400",
  "error": "Bad Request",
  "message": "body should be object"
}
```
I fixed that by explicitly defining the content-type of the request as `application/json`. Then executed the method again and received another error.
```json
{
  "statusCode": "400",
  "error": "Bad Request",
  "message": "body should have required property 'name'"
}
```
Following the link for the [API documentation](https://supabase.github.io/storage-api/#/bucket/post_bucket_) at the [GitHub repo](https://github.com/supabase/storage-api) of Supabase storage middleware I could confirm that `name` and not `id` is the only required property in order to create a new bucket.

I changed the property name and finally manage to create the bucket. 🙂

PR #66, opened 2021-10-15
fix: create_bucket method

julianolf:
- Explicitly define the request content type as `application/json`.
- Pass the bucket name in the `name` attribute instead of `id`.
- Improve tests for bucket operations.
  - Still skipping this test case due missing permissions on test instance.

This PR solves https://github.com/supabase-community/supabase-py/issues/65

PS.: I used my personal account to validate the tests before skipping them.


anand2312:
This has already been changed in #61 , along with adding the `public` parameter to the `create_bucket` method

PR #76, opened 2021-10-30
Remove wrong return type hinting

dreinon:
In line 101, method `from_` returns a Postgrest-py RequestBuilder, not a SupabaseQueryBuilder


J0 (approved):
Looks Good To Me

PR #81, opened 2021-11-14
fixing pypi badge in readme

Phillackinger:
using the right badge "supabase" instad of "supabase-py"


dreinon:
thanks @Phillackinger !

PR #101, opened 2021-12-18
feat: use directly sync postgrest client and remove unused code

### LINKED ISSUE #54: Add support for synchronous RPC calls
RPC calls are asynchronous as `_execute_monkey_patch` doesn't effect them. This is annoying because all the other calls in supabase-py are synchronous.

leynier:
Resolve #54 using directly sync postgrest client and remove unused code


examples/.gitkeep
```diff
@@ -1 +0,0 @@
-
```
J0, line 1: think we might need the `.gitkeep` file in order for the directory to remain in the repo since it has no files/subdirectories.

leynier: The file is not being deleted, what I did was delete an unnecessary line break, as it had two line breaks.

J0: Ah yeah sorry my bad just noticed. Thanks!

supabase/client.py
```diff
@@ -80,21 +78,14 @@ def table(self, table_name: str) -> RequestBuilder:
...
-            table=table_name,
-        )
-        return query_builder.from_(table_name)
+        return self.postgrest.from_(table_name)
 
-    def rpc(self, fn: str, params: Dict[Any, Any]) -> Coroutine[Any, Any, Response]:
+    def rpc(self, fn: str, params: Dict[Any, Any]) -> Response:
         """Performs a stored procedure call.
 
         Parameters
```
dreinon, line 91:
```py
 #     async def remove_subscription_helper(resolve):
    #         try:
    #             await self._close_subscription(subscription)
    #             open_subscriptions = len(self.get_subscriptions())
    #             if not open_subscriptions:
    #                 error = await self.realtime.disconnect()
    #                 if error:
    #                     return {"error": None, "data": { open_subscriptions}}
    #         except Exception as e:
    #             raise e
    #     return remove_subscription_helper(subscription)
```

@J0 why is this code here commented?

supabase/client.py
```diff
@@ -61,7 +59,7 @@ def __init__(
         #     supabase_key=self.supabase_key,
```
dreinon, line 59: Is this comment resolved too?

supabase/client.py
J0: I think at the time it was written `realtime-py` was not ready for release(still isn't but hopefully soon!) As such we commented it out. I think prior to that we may have had a drop in solution which allowed it to work which is why we left it there but I can't quite remember.

Makefile
```diff
@@ -0,0 +1,18 @@
+install:
```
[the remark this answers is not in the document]
J0: This was new to me so I looked it up. In case anyone else is wondering: https://python-poetry.org/docs/pyproject#scripts

supabase/client.py
dreinon: @j0

Makefile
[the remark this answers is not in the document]
dreinon: We can address this in another pr

supabase/client.py
J0: Hey nope I don't think it is, we need to implement realitme in order for it to be resolved

anand2312 (approved):

PR #108, opened 2021-12-29
Update Files For new release

J0:
Note: v0.1.0 because we added sync/async which is a substantial change in the interface? Not sure about this


leynier:
Yeah, v0.1.0 is good for me because the changes of GoTrue Client are breaking changes.

leynier:
I will update to gotrue v0.3.0 and fix the wrap

leynier (approved):

leynier (approved):

PR #110, opened 2021-12-29
fix: update gotrue version and modify client options class

leynier:
Now client options class does not make a deep copy in  the replace method because local storage is an abstract class and not dict like before

### DIFF
--- a/supabase/lib/client_options.py
+++ b/supabase/lib/client_options.py
@@ -1,13 +1,14 @@
-import copy
-import dataclasses
+from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, Optional
+
+from gotrue import SyncMemoryStorage, SyncSupportedStorage
 
 from supabase import __version__
 
 DEFAULT_HEADERS = {"X-Client-Info": f"supabase-py/{__version__}"}
 
 
-@dataclasses.dataclass
+@dataclass
 class ClientOptions:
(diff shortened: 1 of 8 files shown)


J0 (approved):

supabase/lib/client_options.py
```diff
@@ -42,17 +40,19 @@ def replace(
...
+        client_options = ClientOptions()
+        client_options.schema = schema or self.schema
+        client_options.headers = headers or self.headers
+        client_options.auto_refresh_token = (
+            auto_refresh_token or self.auto_refresh_token
+        )
+        client_options.persist_session = persist_session or self.persist_session
+        client_options.local_storage = local_storage or self.local_storage
+        client_options.realtime = realtime or self.realtime
+        client_options.fetch = fetch or self.fetch
```
dreinon, line 57: It would be nice to have a way to do this as a batch, otherwise this will increase in size for every property we add. I don't exactly know how or if it's possible though, but it's worth investigating a bit, don't you think @leynier?

dreinon: Closing this since it works and we want to release soon. Let's investigate it and later on open a new PR to improve this.

dreinon (approved):

dreinon:
Are tests okay?

anand2312 (approved):
LGTM

PR #137, opened 2022-01-30
tests: move subclients tests to subclients

dreinon:


leynier (approved):

anand2312 (approved):

leynier (approved):

tests/conftest.py
```diff
@@ -3,10 +3,15 @@
 import os
 
 import pytest
+from dotenv import load_dotenv
 
 from supabase import Client, create_client
 
 
+def pytest_configure(config) -> None:
```
J0, line 11: This should come in handy :D

tests/test_storage.py
```diff
@@ -0,0 +1,110 @@
...
+            if bucket.id.startswith("pytest-"):
+                storage_client.empty_bucket(bucket.id)
+                storage_client.delete_bucket(bucket.id)
+
+    request.addfinalizer(finalizer)
+
+
+@pytest.fixture(scope="module")
+def bucket(storage_client: SupabaseStorageClient) -> str:
+    """Creates a test bucket which will be used in the whole storage tests run and deleted at the end"""
```
J0, line 39: /nit Was slightly confused by the 8 when I first saw it. Perhaps we could put a comment or a constant somewhere?

J0 (approved):

tests/test_storage.py
dreinon: Sure! We can make a method get_uuid that unifies this. It's a common practice since the uuid type 4 is structured into different parts joined by hyphens, and the first part has a lenght of 8 digits, so it's a way to get a 8 digits long unique identifier that doesn't use hyphens.

anand2312 (approved):

tests/test_storage.py
dreinon: Okay so this wasn't as I explained in the last message. The hex method returns the whole uuid directly without hypens, what returns the uuid with hyphens is  `str(uuid.uuid4())`. Therefore, the only purpose of getting 8 digits is to make the uuid shorter.

PR #140, opened 2022-02-03
tests: ignore 404 when double-checking bucket deletion

dreinon:
In the last PR we introduced some changes to the tests. One of them was a cleanup function that double-checked all test buckets were deleted after the tests run, since the empty and delete methods of the bucket fixture don't run if the test that created the bucket fails.
Nevertheless, tests were still failing because of two reasons:
1. The cleanup function called list_buckets before the buckets were deleted, so sometimes we face a 404 response, which we have to avoid since that means the bucket has already been deleted.
2. Since we were deleting buckets based on the prefix of their id, when multiple test runs are happening at the same time and one finishes before another one, it would delete all test buckets and tests would fail in the runs that were still going .

The solution to 1) is an easy try and except clause continuing the loop if the status code of the response in the StorageException is a 404.
The solution to 2) is using a list as a global variable to only track the buckets that have been created by the corresponding tests run, in order to delete these in the double_check.

### DIFF
--- a/supabase/__init__.py
+++ b/supabase/__init__.py
@@ -4,4 +4,5 @@ from supabase import client, lib
 from supabase.client import Client, create_client
 from supabase.lib.auth_client import SupabaseAuthClient
 from supabase.lib.realtime_client import SupabaseRealtimeClient
-from supabase.lib.storage_client import StorageFileAPI, SupabaseStorageClient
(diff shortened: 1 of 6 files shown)


leynier (approved):

J0 (approved):
Looks Good To Me

PR #167, opened 2022-03-13
feat: upgrade postgrest-py from 0.9.1 to 0.10.0

parrotmac:


pyproject.toml
```diff
@@ -16,7 +16,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-postgrest-py = ">=0.9.1,<0.10.0"
+postgrest-py = ">=0.10.0,<0.11.0"
```
parrotmac, line 19: Is there a reason the version spec has an upper bound?

leynier:
The [poetry.lock](https://github.com/supabase-community/supabase-py/blob/develop/poetry.lock) should be updated also.

leynier:
@parrotmac I'm closing this PR because dependabot already done the job, thanks for the contributions.

PR #179, opened 2022-03-31
Add fns

J0:


supabase/client.py
```diff
@@ -63,6 +74,9 @@ def __init__(
             schema=options.schema,
         )
 
+    def functions(self):
```
anand2312, line 77:
```suggestion
    def functions(self) -> FunctionsClient:
```

anand2312 (changes requested):
LGTM except for the small missing type hint

anand2312 (approved):

PR #218, opened 2022-06-03
fix: pass schema to postgrest initialization 

### LINKED ISSUE #217: Client does not use alternative schema when provided
**Describe the bug**
Client does not pass schema to the initialization of the SyncPostgresClient object. This causes any call to the database to assume that the default public database is used instead of the provided schema. 

**To Reproduce**
1. Create a table in an alternate schema, where a table does not exist with the same name in the public schema.
2. Create a client object using the create_client() function and provide it an alternative schema in the ClientOptions dict.
3. Query the table or attempt an insert into the table

**Expected behavior**
Table is queried / inserted into properly.

**Screenshots**
image

supabase-py version 0.5.6

**Available Workaround**
It is possible to currently call the schema function on the postgrest object within the client object in order to properly set the schema. 
ex: supabase.postgrest.schema('alternateSchema')

Morioki:
Fixes #217

The code was modified to pass options.schema to the _init_postgrest_client function through a new schema parameter to in turn pass the schema information to the SyncPostgrestClient constructor.

### DIFF
--- a/supabase/client.py
+++ b/supabase/client.py
@@ -58,7 +58,7 @@ class Client:
             rest_url=self.rest_url,
             supabase_key=self.supabase_key,
             headers=options.headers,
-            schema=options.schema
+            schema=options.schema,
         )
 
     def storage(self) -> SupabaseStorageClient:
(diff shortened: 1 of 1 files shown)


anand2312 (approved):
LGTM, thanks for the quick fix

anand2312:
@Morioki can you format the file with black as well? CI pre-commit is complaining
I'll merge once CI passes

Morioki:
@anand2312, apologies for not running the formatter before the initial commit. Formatting has now been applied to the client.py using black.

PR #220, opened 2022-06-07
fix: bump storage

dreinon:


anand2312 (approved):
We don't need to do this manually, dependabot should do this automatically haha

PR #287, opened 2022-10-12
Replaced makefile with poetry scripts

### LINKED ISSUE #102: Replace makefile for poetry scripts
This would be more integrated with the poetry ecosystem as well as it would enhance compatibility w windows

cadnce:
Replaced makefile with poetry scripts.

@J0 , @dreinon  is this something like you had in mind?   I'm not really a fan of how you end up with the `poetry_scripts.py` at the top level. But it seems like there isn't really a consensus on how you should do this - https://github.com/python-poetry/poetry/issues/2496

LMK if you had envisaged it somewhat differently/had a better example.

Fixes #102


cadnce:
Oops, fixed that up. Not sure how i missed that. Must have been when I swapped between using poe and the builtin scripts

J0 (approved):
Looks good to me, thanks for the PR!

PR #292, opened 2022-10-22
Async Client support & more

rawandahmad698:

- [x] Deprecates create_client
     A Supabase client can be initialized by directly invoking the class, unlike the JS lib
- [x] Renames `Client` to `SupabaseClient`
- [x] Added async support (thanks to async postgrest-py)
- [x] Moved custom exceptions to `exceptions.py`
- [x] Rewrites README.md with typo, grammatical fixes, and includes async-related code, removes deprecated code

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ We can then read the keys in the python source code.
 import os
 from supabase import SupabaseClient
 
-url: str = os.environ.get("SUPABASE_URL") 
+url: str = os.environ.get("SUPABASE_URL")
 key: str = os.environ.get("SUPABASE_KEY")
 
 supabase: SupabaseClient = SupabaseClient(url, key)
@@ -98,7 +98,7 @@ This is a sample of how you'd use supabase-py. Functions and tests are WIP
 
 ```python
 import os
-from supabase import SupabaseClient 
+from supabase import SupabaseClient
 
 url: str = os.environ.get("SUPABASE_TEST_URL")
 key: str = os.environ.get("SUPABASE_TEST_KEY")
@@ -115,7 +115,7 @@ user = supabase.auth.sign_up(email=random_email, password=random_password)
 
 ```python
 import os
-from supabase import SupabaseClient 
+from supabase import SupabaseClient
 
 
 url: str = os.environ.get("SUPABASE_TEST_URL")
@@ -134,7 +134,7 @@ user = supabase.auth.sign_in(email=random_email, password=random_password)
 
 ```python
 import os
-from supabase import SupabaseClient 
+from supabase import SupabaseClient
 
 
 url: str = os.environ.get("SUPABASE_TEST_URL")
@@ -148,11 +148,11 @@ data = supabase.table("countries").insert({"name":"Germany"}).execute()
 ```python
 async def main():
     import os
-    from supabase import AsyncSupabaseClient 
+    from supabase import AsyncSupabaseClient
 
     url: str = os.environ.get("SUPABASE_TEST_URL")
     key: str = os.environ.get("SUPABASE_TEST_KEY")
-    
+
     async with AsyncSupabaseClient(url, key) as supabase:
         await supabase.table("countries").insert({"name":"Germany"}).execute()
 
@@ -165,7 +165,7 @@ if __name__ == "__main__":
 
 ```python
 import os
-from supabase import SupabaseClient 
+from supabase import SupabaseClient
 
 url: str = os.environ.get("SUPABASE_TEST_URL")
 key: str = os.environ.get("SUPABASE_TEST_KEY")
@@ -175,15 +175,15 @@ supabase: SupabaseClient = SupabaseClient(url, key)
 data = supabase.table("countries").select("*").execute()
 ```
 #### (NEW) Async Support
-    
+
 ```python
 async def main():
     import os
-    from supabase import AsyncSupabaseClient 
+    from supabase import AsyncSupabaseClient
 
     url: str = os.environ.get("SUPABASE_TEST_URL")
     key: str = os.environ.get("SUPABASE_TEST_KEY")
-    
+
     async with AsyncSupabaseClient(url, key) as supabase:
         await supabase.table("countries").select("*").execute()
 
@@ -195,7 +195,7 @@ if __name__ == "__main__":
 
 ```python
 import os
-from supabase import SupabaseClient 
+from supabase import SupabaseClient
 
 url: str = os.environ.get("SUPABASE_TEST_URL")
 key: str = os.environ.get("SUPABASE_TEST_KEY")
@@ -204,15 +204,15 @@ supabase: SupabaseClient = SupabaseClient(url, key)
 data = supabase.table("countries").update({"country": "Indonesia", "capital_city": "Jakarta"}).eq("id", 1).execute()
 ```
 #### (NEW) Async Support
-    
+
 ```python
 async def main():
     import os
-    from supabase import AsyncSupabaseClient 
+    from supabase import AsyncSupabaseClient
 
     url: str = os.environ.get("SUPABASE_TEST_URL")
     key: str = os.environ.get("SUPABASE_TEST_KEY")
-    
+
     async with AsyncSupabaseClient(url, key) as supabase:
         await supabase.table("countries").update({"country": "Indonesia", "capital_city": "Jakarta"}).eq("id", 1).execute()
 
@@ -225,7 +225,7 @@ if __name__ == "__main__":
 
 ```python
 import os
-from supabase import SupabaseClient 
+from supabase import SupabaseClient
 
 url: str = os.environ.get("SUPABASE_TEST_URL")
 key: str = os.environ.get("SUPABASE_TEST_KEY")
@@ -234,18 +234,18 @@ supabase: SupabaseClient = SupabaseClient(url, key)
 data = supabase.table("countries").delete().eq("id", 1).execute()
 ```
 #### (NEW) Async Support
-    
+
 ```python
 async def main():
     import os
-    from supabase import AsyncSupabaseClient 
+    from supabase import AsyncSupabaseClient
 
     url: str = os.environ.get("SUPABASE_TEST_URL")
     key: str = os.environ.get("SUPABASE_TEST_KEY")
-    
+
     async with AsyncSupabaseClient(url, key) as supabase:
         await supabase.table("countries").delete().eq("id", 1).execute()
-        
+
 if __name__ == "__main__":
     import asyncio
     asyncio.run(main())
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ We are currently in Public Alpha. Watch "releases" of this repo to get notified 
 
 ## Installation
 
-**Recommended:** First activate your virtual environment, with your favourites' system. For example, we like `poetry` and `conda`!
+**Recommended:** First, activate your virtual environment with your favourite environment management tool. For example, we like `poetry` and `conda`!
 
 ### PyPi installation
 
--- a/README.md
+++ b/README.md
@@ -143,6 +143,23 @@ supabase: SupabaseClient = SupabaseClient(url, key)
 
 data = supabase.table("countries").insert({"name":"Germany"}).execute()
 ```
+#### (NEW) Async Support
+
+```python
+async def main():
+    import os
+    from supabase import AsyncSupabaseClient 
+
+    url: str = os.environ.get("SUPABASE_TEST_URL")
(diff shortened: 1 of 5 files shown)


rawandahmad698:
and, again i forgot to run `poetry run pre-commit run --all-files` so initial checks will fail

rawandahmad698:
@J0 Hey, could you take a look at this? Also maybe I can DM you and become a listed maintainer while I'm working on adding realtime next.

README.md
```diff
@@ -27,7 +27,7 @@ We are currently in Public Alpha. Watch "releases" of this repo to get notified
 
 ## Installation
 
-**Recomended:** First activate your virtual environment, with your favourites system. For example, we like `poetry` and `conda`!
+**Recommended:** First activate your virtual environment, with your favourites' system. For example, we like `poetry` and `conda`!
```
J0, line 30: /nit Should we rephrase this instead?

```suggestion
**Recommended:** First activate your virtual environment, with your favourite environment management tool. For example, we like `poetry` and `conda`!
```

README.md
```diff
@@ -47,7 +47,7 @@ pip install -e .
 
 ## Usage
 
-It's usually best practice to set your api key environment variables in some way that version control doesn't track them, e.g don't put them in your python modules! Set the key and url for the supabase instance in the shell, or better yet, use a dotenv file. Heres how to set the variables in the shell.
+It's usually best practice to set your API key environment variables in some way that version control doesn't track them, e.g. don't put them in your python modules! Set the Key and URL for the supabase instance in the shell, or better yet, use a dotenv file. Here's how to set the variables in the shell.
```
J0, line 50: /nit Perhaps can consider rephrasing here too
```suggestion
It's usually best practice to set your API key environment variables in some way that isn't tracked by version control. For instance, don't put them in your python modules! Set the Key and URL for the supabase instance in the shell, or better yet, use a dotenv file. Here's how to set the variables in the shell.
```

README.md
```diff
@@ -58,21 +58,22 @@ We can then read the keys in the python source code.
...
 Use the supabase client to interface with your database.
 
 ### Running Tests
 
-Currently the test suites are in a state of flux. We are expanding our clients tests to ensure things are working, and for now can connect to this test instance, that is populated with the following table:
+Currently, the test suites are in a state of flux. We are expanding our clients tests to ensure things are working, and for now can connect to this test instance, that is populated with the following table:
 
 <p align="center">
-  <img width="720" height="481" src="https://i.ibb.co/Bq7Kdty/db.png">
+  <img width="720" height="481" src="https://i.ibb.co/Bq7Kdty/db.png" alt="">
```
J0, line 76: /nit maybe we should fill the alt tag?
```suggestion
  <img width="720" height="481" src="https://i.ibb.co/Bq7Kdty/db.png" alt="example of database table in Supabase Studio editor">
```

README.md
```diff
@@ -96,80 +97,159 @@ This is a sample of how you'd use supabase-py. Functions and tests are WIP
...
 url: str = os.environ.get("SUPABASE_TEST_URL")
 key: str = os.environ.get("SUPABASE_TEST_KEY")
-supabase: Client = create_client(url, key)
+
+supabase: SupabaseClient = SupabaseClient(url, key)
+
 # Create a random user login email and password.
-random_email: str = "3hf82fijf92@supamail.com"
-random_password: str = "fqj13bnf2hiu23h"
+random_email: str = "user@supamail.com"
```
J0, line 109: Hm is there any significant benefit that we gain from changing the email and password in the test examples?

supabase/async_client.py
```diff
@@ -0,0 +1,186 @@
...
+from postgrest.constants import DEFAULT_POSTGREST_CLIENT_TIMEOUT
+from supafunc import FunctionsClient
+
+from .exceptions import SupabaseException
+from .lib.auth_client import SupabaseAuthClient
+from .lib.client_options import ClientOptions
+from .lib.storage_client import SupabaseStorageClient
+
+
+class AsyncSupabaseClient:
```
J0, line 19: Am wondering if we should do away with the sync binding and generate it via `unasync` like how it's done in `Postgrest-py` and `gotrue-py` I think @anand2312 has more expertise here so will let him chime in.

J0 (commented):
Thanks for the PR! Looks like a great start. Left some nits - just have a few concerns:

1. Moving away from `create_client` to `SupabaseClient` -- this should be fine but let me check with the Supabase team if they have any concerns here as there was some past discussion around standardising client libraries.

2. `async` interface seems fine but not sure if we want to make use of [unasync](https://github.com/python-trio/unasync) like in other libraries. Will defer to other maintainers for this one

leynier:
> Thanks for the PR! Looks like a great start. Left some nits - just have a few concerns:
> 
> 1. Moving away from `create_client` to `SupabaseClient` -- this should be fine but let me check with the Supabase team if they have any concerns here as there was some past discussion around standardising client libraries. Do you mind briefly listing the advantages and disadvantages of using `create_client` as opposed to `SupabaseClient`?
> 2. `async` interface seems fine but not sure if we want to make use of [unasync](https://github.com/python-trio/unasync) like in other libraries. Will defer to other maintainers for this one

About the first point, I'm against removing the `create_client` method, because at the moment the only implementation was **synchronous** but by adding the **async** option it may be necessary (if not necessary at the moment, in the future maybe) to create a `create_async_client` method that creates an instance of `AsyncSupabaseClient` **asynchronously** and a `create_sync_client` method that instantiates `SyncSupabaseClient` **synchronously**, and in the case of `AsyncSupabaseClient` instantiation (the **"class constructor"**) is bound to be **synchronous**, so the solution in case an **asynchronous** process is required, such as calling an **asynchronous** **"initialize"** method, could be wrapped in the method `create_async_client`.

anand2312:
I don't see any problems with having the create_client method as we do right now, and I like to not fix things if they aren't broken for the time being 😄 What is your rationale behind this change?

README.md
rawandahmad698: Since these are dummy values, readability is important. I personally think it's easier to read "user@supabase.com" over the random one.

rawandahmad698:
> > Thanks for the PR! Looks like a great start. Left some nits - just have a few concerns:
> 
> > 
> 
> > 1. Moving away from `create_client` to `SupabaseClient` -- this should be fine but let me check with the Supabase team if they have any concerns here as there was some past discussion around standardising client libraries. Do you mind briefly listing the advantages and disadvantages of using `create_client` as opposed to `SupabaseClient`?
> 
> > 2. `async` interface seems fine but not sure if we want to make use of [unasync](https://github.com/python-trio/unasync) like in other libraries. Will defer to other maintainers for this one
> 
> 
> 
> About the first point, I'm against removing the `create_client` method, because at the moment the only implementation was **synchronous** but by adding the **async** option it may be necessary (if not necessary at the moment, in the future maybe) to create a `create_async_client` method that creates an instance of `AsyncSupabaseClient` **asynchronously** and a `create_sync_client` method that instantiates `SyncSupabaseClient` **synchronously**, and in the case of `AsyncSupabaseClient` instantiation (the **"class constructor"**) is bound to be **synchronous**, so the solution in case an **asynchronous** process is required, such as calling an **asynchronous** **"initialize"** method, could be wrapped in the method `create_async_client`.

Not necessarily, a method to invoke an async class is not required, all you need is a base wrapper to initialize `async def __ainit__(self):`

See: 
[Async-Class on PyPi](https://pypi.org/project/async-class/)

supabase/async_client.py
anand2312: Yeah we'd definitely want to generate the sync client with unasync, but I don't think `realtime` is at a useable enough state for us to guarantee a sync interface which would make it difficult to generate two clients here (imo realtime should only have an async interface so only an AsyncSupabaseClient should have realtime features but that's upto debate)

rawandahmad698: If this PR is approved, I’ll be adding realtime to the async class. This was initially my intention.

supabase/async_client.py
```diff
@@ -0,0 +1,186 @@
...
+            supabase_key=self.supabase_key,
+            headers=options.headers,
+            schema=options.schema,
+        )
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        await self.postgrest.session.aclose()
```
anand2312, line 99: 1) We've to do all cleanup here, including closing the other clients as well (ex: `storage`)
2) You can use the [aclose](https://postgrest-py.readthedocs.io/en/latest/api/client.html#postgrest.AsyncPostgrestClient.aclose) method on the postgrest client here
```suggestion
        await self.postgrest.aclose()
```

0xDeadcell:
> > I don't see any problems with having the create_client method as we do right now, and I like to not fix things if they aren't broken for the time being 😄 What is your rationale behind this change?
> 
> 
> I quote you, [The Zen of Python](https://peps.python.org/pep-0020/)
> 
> > Beautiful is better than ugly.
> Explicit is better than implicit.
> Simple is better than complex.
> Complex is better than complicated.
> Flat is better than nested.
> Sparse is better than dense.
> Readability counts.
> 
> Seeing that the create_client is derived from the JS lib, It's extra in Python, and provides zero value and lessens the readability, and simplicity of supabase-py, Invoking the class is much simpler, and easier to use. 
> 
> In the case of async, we can add a async base wrapper to the actual class and use __ainit__;
> 
> Even changing/deprecating the method, or switching to class initialization, won't break anything because there is in fact, nothing special about the method. 
> 
> My problem is that English is not my native language, and I suspect that my explanation is not enough, but I hope a proved a point here. We can choose simplicity, not complexity. 

I agree with you, I don't see why there is any reason not to easily allow async. Hopefully someone can take a look at this again @anand2312

bitnom:
I'm a new Supabase user needing this and don't want to go 140 commits behind the main repo to get it. Someone please resolve it and merge asyncio support.

rawandahmad698:
I sort of "gave up" working with supabase when i was working with realtime (and better async support) and messaged their team for something like a "sandbox" account so i could run my tests without paying anything. The agent said that "It was not their problem" and "They can't help me" even though i don't have a use for supabase myself. Just giving back to the community

NeoFoxxo:
If anyone is still waiting for this to get merged, just clone @rawandahmad698 's repo and downgrade postgrest-py to ``0.10.3`` httpx to ``0.23.0`` and pydantic to ``1.10.2``. I followed the readme and it works perfectly.

J0:
Hey @rawandahmad698,

I'm sorry to hear about your experience with seeking a sandbox account 

> The agent said that "It was not their problem" and "They can't help me" even though i don't have a use for supabase myself.

Do you mind sharing the relevant Discord or email thread where the exact words were used? We'll look into this and see how we can improve the support experience.

One of the features I personally like about Supabase is that it is that the stack is fully Open Source - it's possible to [host supabase locally](https://github.com/supabase/supabase/tree/master/docker) without the need for a hosted instance.  Let me know if you run into any issues getting that set up and I'd be happy to jump in.


With regards to the PR, we'll look into getting this PR updated and re-reviewed this week. Thanks for the contribution again!

Atticuszz:
Hey there! 👋

I've just released an async Supabase client for FastAPI and I'm looking for contributors to help it grow. If you're interested in collaborating on an open-source project and making an impact, your contributions are more than welcome!

Check out the repo and feel free to jump into the issues or discussions. Let's make something great together. 😃

Happy coding!

Atticuszz:
Hey everyone! 🌟 I've been diving deep into the community version of Supabase py and noticed a few areas that could use some tweaks. To address this, I've made some adjustments in my own repository, and I'd love for you to give it a try! Especially designed with FastAPI in mind, this revised version aims to enhance efficiency and compatibility.You know how Supabase js is crafted for frontend use, right? Well, I realized that the existing Supabase py is essentially a mirror of this js version, which doesn't quite fit the bill for Python web frameworks like FastAPI. In FastAPI, every request requires client instances, but I believe this approach isn't always necessary.My proposal? Let's introduce an 'auth clients dict' within the main async client. This way, we won't need to spawn excessive async clients. Instead, we can focus on auth clients, performing operations on Supabase through functions in the main async client (functions that will use the authclient with all the necessary info).I'm excited to collaborate and solve any bugs that might pop up. Let's work together to make the Supabase community even better! 😊 Feel free to join the discussion on GitHub and share your thoughts and suggestions!

silentworks:
Closing this out as async support has been added to the library now. Here is an example below showing how to use the async client.

```python
from supabase._async.client import AsyncClient as Client, create_client

async def create_supabase() -> Client:
    return await create_client(url, key)
```

PR #295, opened 2022-10-25
chore: Update outdated LICENSE year

bluzir:

Updated the outdated copyright year to the present.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Joel Lee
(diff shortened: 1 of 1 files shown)


anand2312:
Hm I'm not a legal expert so I'm not sure how this stuff works, but my [first search result](https://opensource.stackexchange.com/a/1523) seems to suggest that changing the year is not necessary. I'll wait for someone with more experience to comment

J0:
@anand2312 is right, changing the year isn't necessary, or at least not critical at least from a compliance perspective

PR #335, opened 2022-12-29
Update client.py

s360darrius:
The current function url for the official Supabase client is incorrect. I am fixing it by modifying the self.function_url to the correct rpc endpoint.

Bug fix

Currently for supabase environments that end in .co the incorrect function url is used. This results in being unable to utilize supabase functions.

## What is the new behavior?
Corrected the url used so that functions can be used.

## Additional context

### DIFF
--- a/supabase/client.py
+++ b/supabase/client.py
@@ -66,7 +66,8 @@ class Client:
         if is_platform:
             url_parts = supabase_url.split(".")
             self.functions_url = (
-                f"{url_parts[0]}.functions.{url_parts[1]}.{url_parts[2]}"
+                f"{supabase_url}/rest/v1/rpc"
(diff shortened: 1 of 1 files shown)


J0:
Hey @s360darrius,

Thanks for the PR! The functions url here refers to [Edge functions](https://supabase.com/docs/guides/functions) rather than [Database Functions / RPC](https://supabase.com/docs/guides/database/functions)


To make use of Database functions one would call the [`.rpc()` function](https://github.com/supabase-community/postgrest-py/blob/master/postgrest/_async/client.py#L79) on `postgrest-py`

J0:
Going to close this for now but feel free to let us know if there are further questions

Issue #185, opened 2022-04-11
RLS not working after sign-in


tobias-scheepers:
**Describe the bug**
After signing in a user using ` supabase-py` a user is correctly returned with it's acces_token etc. However if I try to insert or select a row in a table with RLS enabled it seems the client does not correctly send the requests as being authenticated by that user. Only the tables without RLS seem to be working. The error returned by supabase is:

```
postgrest_py.exceptions.APIError: {'message': 'new row violates row-level security policy for table "machines"', 'code': '42501', 'details': None, 'hint': None}
```

Is this currently not supported by supabase-py or is this indeed a bug? Also, if not supported how would one best go about accessing secured tables (e.g. using postgres-py directly).

**To Reproduce**
The minimal example below highlights this for systems 

```
client = create_client(SUPABASE_URL, SUPABASE_KEY)
session = client.auth.sign_in(email=SUPABASE_MAIL, password=SUPABASE_PASS)
user = session.user

data = {
    "user_id": str(user.id)
}

response = client.table("examples").insert(data).execute()

client.auth.sign_out()
```

anand2312:
Hi, using the signed-in user's bearer token has not yet been implemented. I am planning on getting to this soon-ish.

tobias-scheepers:
Thanks @anand2312 this clears things up indeed. The workaround works flawlessly in the mean time!

timreibe:
+1, coming from #272

anand2312:
@timreibe 👍🏼 done. Perhaps this issue should be pinned for better visibilty

nielsrolf:
Are there any updates on this? :)

J0:
Hey @nielsrolf,

Thanks for the query! Seems like this is related to #272 - let me pin the issue as well

0xDeadcell:
Any updates on this?

0xDeadcell:
> This needs some work to be done on postgrest-py first (as in, it isn't directly supported as such). I'm working on that right now, and it should be done by the end of this week, and when that's done we'll make a new release of postgrest-py and have it work here.
> 
> In the meantime you could do something along the lines of:
> 
> ```python
> client = create_client(SUPABASE_URL, SUPABASE_KEY)
> session = client.auth.sign_in(email, password)
> 
> postgrest_client = client.postgrest
> postgrest_client.auth(session.access_token) 
> 
> # run queries
> client.auth.sign_out()
[...]

PR #370, opened 2023-02-11
fix: actually fetch access token from session

### LINKED ISSUE #58: Row Level Security and Python
Hi Folks!

I am calling a python API from  Vue app where an user is logged in using supabase. This Python API was created using FastAPI and I have shared the same JWT secret and algorithm so I can ensure who is calling the API. The problem is when I try to execute an insert using supabase client + anon key I am receiving a RLS error, seems that it can't identify who is calling the insert from the python client.

My RLS is:
```
Enable insert for authenticated users only
(role() = 'authenticated'::text)
```

Is it possible somehow to pass which user is calling the python API using supabase-py? I do have the access_token sent by the javascript client but I have no idea how to share it and use on `supabase.table("projects").insert`

J0:
Pass through access token

Aims to fix #58 and #185

### DIFF
--- a/supabase/client.py
+++ b/supabase/client.py
@@ -197,12 +197,13 @@ class Client:
         client.auth(token=supabase_key)
         return client
 
-    def _get_auth_headers(self) -> Dict[str, str]:
+   def _get_auth_headers(self) -> Dict[str, str]:
         """Helper method to get auth headers."""
         # What's the corresponding method to get the token
+        data = self.auth.get_session()
         return {
             "apiKey": self.supabase_key,
-            "Authorization": f"Bearer {self.supabase_key}",
+            "Authorization": f"Bearer {data.get('access_token')}",
         }
 
 


SaileshBechar (approved):

supabase/client.py
```diff
@@ -200,9 +202,11 @@ def _init_postgrest_client(
     def _get_auth_headers(self) -> Dict[str, str]:
         """Helper method to get auth headers."""
         # What's the corresponding method to get the token
+        data = self.auth.get_session()
+        token = data.get("access_token") if data else None
         return {
             "apiKey": self.supabase_key,
-            "Authorization": f"Bearer {self.supabase_key}",
+            "Authorization": f"Bearer {token}",
```
kamrik, line 209: I think this line is executed only when the client is initialized, that is before sign-in. We need something that sets the header not only after sign-in but also every time the access token is refreshed. Maybe we can add a function like `Client.ensure_fresh_session()` that would get the token from `auth.get_session()` (which refreshes access token if needed) and then set the authorization header for all sub-components of the `Client`.

J0: Yup, I agree, there's a refresh that is needed. In the js lib there's a refresh triggered when the auth state changes. I don't quite have bandwidth to look into this outside of weekends but if anyone has bandwidth feel free to pick it up

SaileshBechar: It feels natural for the correct token to be set when calling `Client..auth.set_session()`. Could the headers just be updated as a part of this workflow?

kamrik: Then the typical workflow will be:
1. sign_in
2. set_session()
3. call to retrieve data from the DB
// some time later, when the access token might have expired
4. set_session() again
5. another call to retrieve data from the DB

Forcing people to call `set_session()` every time before DB calls seems redundant. Am I missing something?

We could change the implementation of `Client.from_()` to take care of token refresh and header, then the user can work in a more intuitive way. 
1. sign in
2. call DB

0xDeadcell:
Can we have someone approve this please so this can get merged

vcscsvcscs (approved):
This should be approved and merged, as this is an insanely annoying issue.

0xDeadcell:
Can someone with write access finalize this? @J0 @kamrik @anand2312 Would really appreciate it <3

J0:
Hey everyone,

Sorry for the incommunicado - I don't think the changes in the PR actually fix the issue as there's still a need to properly implement the refresh of the `access_token`. There's someone who is looking into this right now and will also comb through the other issues so do watch this space. They will also do a pass through the library so hopefully that should improve the state we're in right now.

I'm going to go ahead and close this PR for now. In hindsight, I should have done this earlier so sorry about this. For anyone who is looking for how to do access token verification for a Python backend you can consider taking a look at the good work done by  StanGirard: https://github.com/StanGirard/quivr/pull/144

PR #381, opened 2023-02-21
fix: add storage client timeout

J0:


anand2312 (commented):
I think it's nice we're making the storage client also an instance variable on `supabase.Client`, it being accessible through the `.storage()` method while all the others are instance variables was a bit confusing 😄

PR #458, opened 2023-06-12
fix wrong pytest configuration

danhdevelop:

Bug fix, Pytest running with error about importing supabase package


J0 (approved):
LGTM

PR #471, opened 2023-06-17
chore: fixed some types

Ananya2001-an:

Fixed some types and also refactored the client options test by creating a class to make it more readable.

### DIFF
--- a/supabase/client.py
+++ b/supabase/client.py
@@ -59,12 +59,12 @@ class Client:
         self.supabase_url = supabase_url
         self.supabase_key = supabase_key
         options.headers.update(self._get_auth_headers())
-        self.rest_url: str = f"{supabase_url}/rest/v1"
-        self.realtime_url: str = f"{supabase_url}/realtime/v1".replace("http", "ws")
-        self.auth_url: str = f"{supabase_url}/auth/v1"
+        self.rest_url = f"{supabase_url}/rest/v1"
+        self.realtime_url = f"{supabase_url}/realtime/v1".replace("http", "ws")
+        self.auth_url = f"{supabase_url}/auth/v1"
         self.storage_url = f"{supabase_url}/storage/v1"
         self.functions_url = f"{supabase_url}/functions/v1"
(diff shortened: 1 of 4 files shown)


supabase/client.py
anand2312, line 1: Not sure how I feel about these changes. The types can be trivially inferred for all those attribute assignments (`pyright` does so locally for me)
I'm not sure why some of them already had the `: str` on there but I feel like it'd be better to remove them altogether - I believe we should only explicitly type-hint when the type-checker cannot infer/the type is ambiguous

tests/test_client_options.py
```diff
@@ -3,41 +3,41 @@
...
-    options = ClientOptions(storage=storage)
-    new_options = options.replace()
-
-    # Act
-    new_options.storage.set_item("key", "new_value")
-
-    # Assert
-    assert options.storage.get_item("key") == "new_value"
-    assert new_options.storage.get_item("key") == "new_value"
+class TestClientOptions:
```
anand2312, line 6: Hmm I don't get the rationale here - can you elaborate how the class makes the tests more readable?

anand2312 (changes requested):
Thanks for the PR! I like the changes made to the `auth_client` type hints but I have some questions about the other changes

tests/test_client_options.py
Ananya2001-an: Both the methods were for testing client options so instead of prefixing the word `test_client_options` again and again it was better to keep both inside a class which broadly describes their purpose.

supabase/client.py
Ananya2001-an: Yeah I guess you are right...I will remove them; it's only useful in case of function parameters :)

tests/test_client_options.py
anand2312: Fair enough I guess

anand2312 (approved):

J0 (approved):
LGTM, thanks for the PR!

PR #481, opened 2023-06-27
fix: remove fetch from clientoptions

anand2312:
Looks like `fetch` was included in the ClientOptions because it must've been copied over from the JS client 😦

### DIFF
--- a/supabase/lib/client_options.py
+++ b/supabase/lib/client_options.py
@@ -34,9 +34,6 @@ class ClientOptions:
     realtime: Optional[Dict[str, Any]] = None
     """Options passed to the realtime-py instance"""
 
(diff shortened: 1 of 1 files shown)


anand2312:
Can we re-run CI here? It's been a while and I'm not sure what this change breaks in the tests

J0 (approved):

J0:
Hm odd, should probably patch the tests but otherwise lgtm

J0:
@anand2312 do you mind pushing an empty commit or similar? Can't figure out how to trigger a re-run since the branch is a little old. Seems like there's only the option to delete logs

anand2312:
Done - looks like pre-commit was complaining about an unused import

PR #489, opened 2023-07-08
Update to support automatically synchronizing access token between libraries

danhdevelop:

Updated to support automatically synchronizing access token between gotrue-py, postgrest-py and storage-py

Developers need to manually get an access token from gotrue-py and pass it to postgrest-py, storage-py... to authorize the connection to the backend.

## What is the new behavior?

All libraries share an http client class which contains a static variable to share authentication information.

## Additional context

Shared access token allow developers to reduce the steps to call API
When the access token is expired, the new token after being refreshed will be automatically supplied to all other libraries.


pyproject.toml
```diff
@@ -23,6 +23,7 @@ httpx = "^0.23.0"
 storage3 = "^0.5.2"
 supafunc = "^0.2.2"
 python-semantic-release = "7.34.6"
+faker = "^18.10.1"
```
anand2312, line 26: I think faker ought to be a dev dependency

anand2312 (commented):
I don't see code for the issue mentioned in the title - just a comment in the meantime

pyproject.toml
danhdevelop: yes, you are right @anand2312 the faker should be at dev dependency.

silentworks:
Thank you for your attempt at fixing this issue. We've merge another PR which solves this issue. https://github.com/supabase-community/supabase-py/pull/560

PR #530, opened 2023-08-30
Attempt to provide a fix for the RLS syncing issue.

Blastorios:

Fix the RLS syncing issue ( #185 ) existing between the independent supabase and postgrest client.

User authentication does not invoke an update in the postgrest-py session, which causes an RLS error to appear as the authentication header remains on the default `supabase_key`.

## What is the new behavior?

If the postgrest-py clients are updated to match:
```python
from typing import Optional

# ...

def from_(self, table: str, auth_token: Optional[str] = None) -> SyncRequestBuilder:
   # Add the auth_token parameter with the default None value 
   # to introduce a non-intrusive update that will allow other clients to sync their auth state.
    if auth_token is not None:
       self.session.headers["Authorization"] = f"Bearer {auth_token}"

    return SyncRequestBuilder(self.session, f"/{table}")

# ...
```

Every `.table()` call from the supabase client will invoke a `_get_auth_token` check that will provide either the `supabase_key` or the new `access_token`. This resembles the javascript implementation as every fetch call will trigger an 'auth check' function to retrieve the available session.

I am uncertain if this will fix the issue for all 3 login-types.

## Additional context

This PR should not be merged before postgrest-py is also fixed (with the proposed snippet and a fixed dependency version).


silentworks:
Thank you for your attempt at fixing this issue. We've gone with a more flexible option which doesn't require passing the auth token to the individual methods and also covers RLS for storage too. https://github.com/supabase-community/supabase-py/pull/560

PR #552, opened 2023-09-14
Added asynchronous methods to supabase client

aschung01:

✅  Added async support to Client
  

### Usage
```
await client.atable("table_name").select("*").execute()
```

### DIFF
--- a/supabase/client.py
+++ b/supabase/client.py
@@ -1,8 +1,15 @@
 import re
-from typing import Any, Dict, Union
+from typing import Any, Dict, Union, Coroutine
 
 from httpx import Timeout
-from postgrest import SyncFilterRequestBuilder, SyncPostgrestClient, SyncRequestBuilder
+from postgrest import (
+    SyncFilterRequestBuilder,
+    SyncPostgrestClient,
+    SyncRequestBuilder,
+    AsyncFilterRequestBuilder,
+    AsyncPostgrestClient,
+    AsyncRequestBuilder,
+)
 from postgrest.constants import DEFAULT_POSTGREST_CLIENT_TIMEOUT
 from storage3.constants import DEFAULT_TIMEOUT as DEFAULT_STORAGE_CLIENT_TIMEOUT
 from supafunc import FunctionsClient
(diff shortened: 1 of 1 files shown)


aschung01:
@J0 Hey Joel, could you please check this out? It would be nice to be able to call both async and synchronous methods from one client.

Wamy-Dev:
I would love async

J0:
Hey we'll take a look shortly (after work)

Could we trouble you to run the corresponding pre-commit hooks?

You can do
`pre-commit install`

and then run something like `pre-commit run --all`

anand2312 (changes requested):
Just skimmed through - but generally our approach for async support is to make two separate clients, one that has only async methods and another that has only sync methods. I'm partial to that approach because you can choose which client to use and only initialise that (here we're now initialising both a sync and async postgrest client in the `__init__` of the supabase client)

Atticuszz:
> I would love async

i published the async package 😊

Atticuszz:
Hey there! 👋

I've just released an async Supabase client for FastAPI and I'm looking for contributors to help it grow. If you're interested in collaborating on an open-source project and making an impact, your contributions are more than welcome!

Check out the repo and feel free to jump into the issues or discussions. Let's make something great together. 😃

Happy coding!

Atticuszz:
Hey everyone! 🌟 I've been diving deep into the community version of Supabase py and noticed a few areas that could use some tweaks. To address this, I've made some adjustments in my own repository, and I'd love for you to give it a try! Especially designed with FastAPI in mind, this revised version aims to enhance efficiency and compatibility.You know how Supabase js is crafted for frontend use, right? Well, I realized that the existing Supabase py is essentially a mirror of this js version, which doesn't quite fit the bill for Python web frameworks like FastAPI. In FastAPI, every request requires client instances, but I believe this approach isn't always necessary.My proposal? Let's introduce an 'auth clients dict' within the main async client. This way, we won't need to spawn excessive async clients. Instead, we can focus on auth clients, performing operations on Supabase through functions in the main async client (functions that will use the authclient with all the necessary info).I'm excited to collaborate and solve any bugs that might pop up. Let's work together to make the Supabase community even better! 😊 Feel free to join the discussion on GitHub and share your thoughts and suggestions!

silentworks:
Closing this out as async support has been added to the library now. Here is an example below showing how to use the async client.

```python
from supabase._async.client import AsyncClient as Client, create_client

async def create_supabase() -> Client:
    return await create_client(url, key)
```

PR #560, opened 2023-09-24
Fix issue of RLS not working with Postgrest and Storage

### LINKED ISSUE #185: RLS not working after sign-in

silentworks:

This will now allow the correct auth headers for a authenticated user to be passed over to the postgrest client. In this setup the postgrest and storage client are lazy initialised and each time a event triggers that could cause an auth token to update we reset the private variable holding the postgrest and storage clients.

No complying with the RLS policies without handling it in your code yourself

## What is the new behavior?

Automatic handling of complying with the RLS polices with postgrest and storage.

## Additional context

Fixes #185

### DIFF
--- a/supabase/client.py
+++ b/supabase/client.py
@@ -80,9 +80,7 @@ class Client:
         # )
         self.realtime = None
         self._postgrest = None
-        self.storage = self._init_storage_client(
-            self.storage_url, self._get_auth_headers(), options.storage_client_timeout
-        )
+        self._storage = None
         self.auth.on_auth_state_change(self._listen_to_auth_events)
 
     def functions(self) -> FunctionsClient:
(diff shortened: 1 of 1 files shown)


J0 (approved):
Lgtm if tested

ShantanuNair:
@silentworks Much appreciated!!

Issue #576, opened 2023-10-02
Change the functions method inside of supabase-py to a property


silentworks:
Currently the supabase-py library has a functions method but this should follow all the other services inside of the library and use a property instead.

To currently invoke a function your code looks like:
```python
supabase.functions().invoke()
```

This change will make this code look like:
```python
supabase.functions.invoke()
```

anand2312:
This will be a breaking change and as we've made a 1.0 release we'd be looking at a major-version change for this (following semver). What do you think our approach should be - allow the `functions().invoke()` with a deprecation warning till a 2.0 release? @silentworks @J0

J0:
> allow the functions().invoke() with a deprecation warning till a 2.0 release (while making functions.invoke() the recommended default)

Sounds good to me

PR #579, opened 2023-10-03
feat: add functions property

J0:

see title, to fully remove in v1.3.0. Address #576 

 - need to update docs


silentworks (changes requested):
You need to also reset the `self._functions` to `None` whenever an event fires inside of
```
def _listen_to_auth_events(self, event: AuthChangeEvent, session):
```

J0:
Added

silentworks (approved):

PR #585, opened 2023-10-10
README / Storage -- Update `delete` file to `remove` file

devinem4:
Swap `delete` out, `remove` in the Storage section of the top-level readme.

Documentation update

Documentation is using a `delete` function to remove a file, but I was hitting an error when using it:
`AttributeError: 'SyncBucketProxy' object has no attribute 'delete'`

I'm pretty sure `remove` is the keyword that we need to use:
https://github.com/supabase-community/storage-py/blob/main/storage3/_sync/file_api.py#L277

## What is the new behavior?

n/a - no functionality changes just documentation

## Additional context

I'm brand new to supabase and really loving this project. Just hope this documentation update helps some other noob out down the line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ new_file = getUserFile()
 data = supabase.storage.from_(bucket_name).upload("/user1/profile.png", new_file)
 ```
 
-### Delete a file
+### Remove a file
 
 ```python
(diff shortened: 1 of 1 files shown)


J0 (approved):
Thanks for the PR!

PR #605, opened 2023-10-29
feat(functions-py): update functions-py version

silentworks:
BREAKING CHANGE: Functions now raise exceptions on errors


J0 (approved):
Lgtm, thanks! Can we have make build_sync at some point in the future too?

silentworks:
@J0 yeah I will try and get async done for `supabase-py` library this week so we can add `make build_sync` back in for this library.

PR #610, opened 2023-11-01
fix: add flow_type to client options

silentworks:

Add flow_type option to client options for gotrue-py

Only supports `implicit` auth flow

## What is the new behavior?

Can now select between `pkce` and `implicit` auth flow

## Additional context


anand2312 (approved):
LGTM

PR #630, opened 2023-11-30
feat: add create method to handle token headers

silentworks:

Add a new create method to handle async token retrieval

The library isn't fully async and complains when trying to call `get_session` inside of `_get_token_header` from a `@property`.

## What is the new behavior?

The library should now be fully async as the `get_session` can be called async since we've moved the `_get_token_header` to the new `create` method instead of calling it from a `@property`.

## Additional context


J0 (approved):
🌶️ 🔥

PR #656, opened 2024-01-07
update typehint and  add optional access_token to creat_client

Atticuszz:
fix the rpc return typehint of async_client and add async typehint to client options
and add the optional access_token to creat_client

as for front end ，no need to pass refresh token every time，just access token，tokens should handled by supabase-js auth
as for auth by supabase-py，same to before


supabase/_async/client.py
```diff
@@ -4,11 +4,8 @@
 from gotrue._async.storage import AsyncMemoryStorage
 from gotrue.types import AuthChangeEvent, Session
 from httpx import Timeout
-from postgrest import (
-    AsyncFilterRequestBuilder,
-    AsyncPostgrestClient,
-    AsyncRequestBuilder,
-)
+from postgrest import AsyncPostgrestClient, AsyncRequestBuilder
+from postgrest._async.request_builder import AsyncRPCFilterRequestBuilder
```
silentworks, line 8: Is this not imported from the `postgrest` namespace itself? if its only available from the `postgrest._async.request_builder` maybe we will need to also make it available from `postgrest` otherwise I don't think `unasync-cli` will work.

silentworks: I've tested this and it breaks, I'm updating the `postgrest` package now to export `AsyncRPCFilterRequestBuilder` from the main package and then this can be updated and merged.

Atticuszz: thanks for great works！

silentworks:
Hey @Atticuszz, not sure why you closed this as I think this is a good PR to get merged in. I've reopened it and will be merging it in soon.

Atticuszz:
@silentworks also i add the optional access_token to creat_client

silentworks:
@Atticuszz I need to investigate the `access_token` issue a bit more first, lets not include it in the PR so I can get this merged in.

Atticuszz:
> @Atticuszz I need to investigate the `access_token` issue a bit more first, lets not include it in the PR so I can get this merged in.

Okay, good luck finding the problem soon！

Atticuszz:
it's too old now,so closed it

Issue #682, opened 2024-02-04
Error importing `ClientOptions` from supabase


dantheman0207:
**Describe the bug**
I get this error: `ImportError: cannot import name 'ClientOptions' from 'supabase'`

**To Reproduce**
This is my code:
```Python
from supabase import create_client, Client, ClientOptions
from .config import Settings

settings = Settings()

def get_supabase(jwt: str = None) -> Client:
    if jwt:
        return create_client(
            settings.SUPABASE_URL,
            settings.SUPABASE_KEY,
            options=ClientOptions(
                headers={"Authorization": f"Bearer {jwt}"}
            )
        )
    else:
        return create_client(settings.SUPABASE_URL, settings.SUPABASE_KEY)
```
**Expected behavior**
This is using the exact code described [in the docs](https://supabase.com/docs/reference/python/initializing). You can see a screenshot of it below. This import should work as documented.

**Screenshots**
<img width="554" alt="Screenshot 2024-02-04 at 15 09 09" src="https://github.com/supabase-community/supabase-py/assets/1239724/910af9f1-b0cc-4ab1-bd74-e015ebbf49ab">

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.2.1
 - Browser: Safari
 - Version: 
    - supabase==2.3.4
    - Python 3.11.7
    - Using virtualenv

**Additional context**
I am trying to set the header to the user's JWT auth token so I can implement RLS policies, as this will prevent me from writing a bunch of backend code to authorize access.

Atticuszz:
```python
async def create(
        cls,
        supabase_url: str,
        supabase_key: str,
        options: ClientOptions = ClientOptions(),
    ):
        client = cls(supabase_url, supabase_key, options)
        client._auth_token = await client._get_token_header()
        return client

    

    @property
    def postgrest(self):
        if self._postgrest is None:
            self.options.headers.update(self._auth_token)
            self._postgrest = self._init_postgrest_client(
                rest_url=self.rest_url,
                headers=self.options.headers,
                schema=self.options.schema,
                timeout=self.options.postgrest_client_timeout,
            )

        return self._postgrest
```
the code in repo
reset the your jwt by `client._auth_token = await client._get_token_header()`，so it occured as you said

dantheman0207:
@Atticuszz I don't want to reset the JWT, I want to use a specific JWT associated with the user using my API (so I can take advantage of RLS policies). 

[...]

PR #688, opened 2024-02-14
fix: add missing ClientOptions to the main init file

silentworks:

Bug fix for #682

Missing `ClientOptions` when importing from the `supabase` namespace

## What is the new behavior?

Adding `ClientOptions` to the `__init__.py` file

## Additional context


J0 (approved):
Thanks!

PR #750, opened 2024-03-26
remove mdformat-gfm

silentworks:

Removing mdformat-gfm as it had no real benefit to the project and is creating friction for the generated CHANGELOG.md

Changelog is failing in formatting tests as it doesn't follow github flavour md.

## What is the new behavior?

Changelog won't failing the formatting tests.

## Additional context


J0:
Thanks! Missed this on review. Agree that we shouldn't keep i it if it's messing with changelog

PR #760, opened 2024-04-11
add import async client

fakelog:
I added a simple import of asynchronous functions similar to synchronous ones

### DIFF
--- a/supabase/__init__.py
+++ b/supabase/__init__.py
@@ -2,15 +2,31 @@ from postgrest import APIError as PostgrestAPIError
 from postgrest import APIResponse as PostgrestAPIResponse
 from storage3.utils import StorageException
 
+# Version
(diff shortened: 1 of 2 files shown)


fakelog:
> Most IDEs will even write the `from ... import ...` line for you.


the problem is the access level

Снимок экрана_20240412_151450

silentworks:
I'm not sure what you mean by access level. This looks like an IDE warning and not a Python runtime warning. Can you clarify this please?

fakelog:
> Can you clarify this please?

python supports access modifiers, although they are not really something strict and the interpreter will not give an error, but they are still there and the IDE highlights this as a warning

async - public
_async - protected
__async - private

silentworks (approved):
Thanks for this PR. This change will be available on the next release.

Issue #762, opened 2024-04-12
Authorization headers silently shared between clients.


hssumar:
# Bug report

## Describe the bug

If you do

```
anon_client = supabase.create_client(SUPABASE_URL, SUPABASE_ANON_KEY)
service_client  = supabase.create_client(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY)
```

Then `anon_client` will have service role privileges, which is pretty bad and could cause security issues. 
E.g. `anon_client.auth.admin.list_users()` will just list all the users.

This happens because thedefault  `SyncMemoryStorage` in 

https://github.com/supabase-community/supabase-py/blob/main/supabase/_sync/client.py#L35

is the same object for both clients, due to https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Hence the authoriztion header with the service role key will enter into the anon client.

silentworks:
Thanks for this. I was able to replicate this and will push out a patch version with a fix soon.

J0:
Thanks for reporting the fix and the follow on PR

silentworks:
This has been addressed in release 2.4.2

PR #764, opened 2024-04-12
fix: remove default mutable argument from client creation

silentworks:

Bug fix

Currently leaking session data if you instantiate two clients in the same request.

## What is the new behavior?

No longer leaking session data.

## Additional context

Issue #762


J0:
Thanks for the quick fix! Tested locally and seems to be working as expected

J0 (approved):

Issue #420, opened 2023-04-24
Create a client with Auth context of a user


vlebert:
Hi everyone

**Is your feature request related to a problem? Please describe.**

I am trying to write python cloud function (instead of supabase edge function). I wan't to get caller's identity do proceed database read/write with his RLS context.
In JS, this is possible as described in the documentation.

https://supabase.com/docs/guides/functions/auth

```js
import { serve } from 'https://deno.land/std@0.177.0/http/server.ts'
import { createClient } from 'https://esm.sh/@supabase/supabase-js@2'

serve(async (req: Request) => {
  try {
    // Create a Supabase client with the Auth context of the logged in user.
    const supabaseClient = createClient(
      // Supabase API URL - env var exported by default.
      Deno.env.get('SUPABASE_URL') ?? '',
      // Supabase API ANON KEY - env var exported by default.
      Deno.env.get('SUPABASE_ANON_KEY') ?? '',
      // Create client with Auth context of the user that called the function.
      // This way your row-level-security (RLS) policies are applied.
      { global: { headers: { Authorization: req.headers.get('Authorization')! } } }
    )
```

With Python client, I couldn't reproduce. I tried:

```python
supa_client = create_client("https://****.supabase.co",
                            "***anon_api_key***",
                            ClientOptions().replace(headers={"authorization":"Bearer ***user_session_token***"
}))
```

I also tried

```python
supa_client = create_client("https://****.supabase.co",
                            "***anon_api_key***",
}))
supa_client.auth.set_session("***user_session_token***","")

```

None of this works. After studying the code a bit, I think this may be the problem:

https://github.com/supabase-community/supabase-py/blob/2bba842449ccd0b5f933198c343f54c5a67db7ed/supabase/client.py#L61

https://github.com/supabase-community/supabase-py/blob/2bba842449ccd0b5f933198c343f54c5a67db7ed/supabase/client.py#L208

Authorization token is always overwritten with anon API KEY

```python
options.headers.update(self._get_auth_headers())
```

```python
def _get_auth_headers(self) -> Dict[str, str]:
        """Helper method to get auth headers."""
        # What's the corresponding method to get the token
        return {
            "apiKey": self.supabase_key,
            "Authorization": f"Bearer {self.supabase_key}",
        }
```

**Describe the solution you'd like**

It should be possible to reproduce JS behavior to create client with Auth context of the user that called the function (logged in user's JWT).

Am I missing something ?

vlebert:
[...]

PR #766, opened 2024-04-13
fix(user auth context): do not overwrite provided client options Authorization header

Garee:

Bug fix for #420 

Supports setting the client auth context for a particular user and therefore RLS.

- Updated both sync and async clients.
- Added unit tests.

If a client is created with an `Authorization` header provided in `ClientOptions`, it is overwritten with the supabase key.

For example:

```python
url = "https://example.supabase.com"
key = "apikey"
options = ClientOptions(headers={"Authorization": "Bearer userjwt"})
supabase = create_client(url, key, options)

# Subsequent requests will include the headers:
# apiKey: "apiKey"
# Authorization: "Bearer apiKey" <- Expected "Bearer userjwt"
```

## What is the new behavior?

The `Authorization` header provided via client options is used for requests.

```python
# apiKey: "apikey"
# Authorization: "Bearer userjwt"
```

If no header is provided the supabase key is used.

## Additional context

This allows us to set the auth context in the same manner [described in the official Edge Function documentation](https://supabase.com/docs/guides/functions/auth#auth-context).

### DIFF
--- a/supabase/_async/client.py
+++ b/supabase/_async/client.py
@@ -64,11 +64,8 @@ class AsyncClient:
 
         self.supabase_url = supabase_url
         self.supabase_key = supabase_key
-        self._auth_token = {
-            "Authorization": f"Bearer {supabase_key}",
-        }
-        options.headers.update(self._get_auth_headers())
         self.options = options
+        options.headers.update(self._get_auth_headers())
         self.rest_url = f"{supabase_url}/rest/v1"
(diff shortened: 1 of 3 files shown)


silentworks:
I'm currently reviewing this PR. I will need to test it out for a bit as some code has been removed which were to fix a previous issue.

silentworks (approved):
Thanks for this PR.

Issue #510, opened 2023-07-28
httpx.ReadTimeout: The read operation timed out


ishaqmahsud:
**Describe the bug**
to call vector search on 73k records supabase (python) failed due timeout error. please put option developer can increase or decrease timeout according situation. otherwise python library not worthful.

**To Reproduce**
the error can be produce to make embedding to search in huge records about 73k.

**Expected behavior**
i need to increate the timeout for the rpc method so it can easily with failed. 
in JavaScript its working where it can takes 25 to 30 secs in python it failed. 

**Screenshots**
Traceback (most recent call last):
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpx\_transports\default.py", line 60, in map_httpcore_exceptions
    yield
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpx\_transports\default.py", line 218, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpcore\_sync\connection_pool.py", line 253, in handle_request
    raise exc
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpcore\_sync\connection_pool.py", line 237, in handle_request
    response = connection.handle_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpcore\_sync\connection.py", line 90, in handle_request
    return self._connection.handle_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpcore\_sync\http11.py", line 112, in handle_request
    raise exc
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpcore\_sync\http11.py", line 91, in handle_request
    ) = self._receive_response_headers(**kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\fielder\textkernelfunc\Lib\site-packages\httpcore\_sync\http11.py", line 155, in _receive_response_headers
    event = self._receive_event(timeout=timeout)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #509, opened 2023-07-28
The operation did not complete (write) (_ssl.c:2396)


Mansitos:
I'm getting error 

**The operation did not complete (write) (_ssl.c:2396)** 

while loading embeddings through langhchain on supabase.

The function that get this error is:
`self.embeddings_db = SupabaseVectorStore.from_documents(self.docs, self.embeddings, client=supabase_client)`

specifically, the error is produced in row: `result = client.from_(table_name).insert(chunk).execute()`
from `supabase.py` file, function `_add_vectors()`

By changing `chunk_size` variable to a smaller value (it is hard-coded to 500) it fixes the problem.

Any suggestion? any way to change chunk_size?

![image](https://github.com/supabase-community/supabase-py/assets/36408034/36c20c2b-a54d-4cfa-be25-c2c9e2a7636a)

J0:
Perhaps it might be timing out as we use `httpx` which has a built in timeout of 5s. Don't think the js client has a timeout - was going to suggest moving the vector store to `vecs` which is an official Supabase-py for embeddings and probably more performant for working with embeddings

Do you mind opening an issue about this on langchain instead? Feel free to tag me if needed

silentworks:
Stale issue so closing this one out.

Issue #487, opened 2023-07-05
Initializing python supabase client with timeout does not work as displayed on website. 


Simar-Kohli:
**Describe the bug**

I'm trying to set a timeout on the client (since i've been getting httpx read timeout errors). 

```
import os
from supabase import create_client, Client

url: str = os.environ.get("SUPABASE_URL")
key: str = os.environ.get("SUPABASE_KEY")
supabase: Client = create_client(url, key, options={'timeout': 10})
```

This is as displayed on the website: https://supabase.com/docs/reference/python/initializing

**To Reproduce**
Put this code into a python file, and attempt to execute with current public pip3 install of supabase. 

**Expected behavior**
I expected it to be a valid client instance and start to use it for my python script. Instead, I get this error: 

Traceback (most recent call last):
  File "/Users/XXXXX/Desktop/XXXXXX/XXXXXX/deployableApprovals.py", line 27, in <module>
    supabaseClient: Client = create_client(SUPABASEURL, SUPABASEKEY, {'timeout': 10})
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/supabase/client.py", line 248, in create_client
    return Client(supabase_url=supabase_url, supabase_key=supabase_key, options=options)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/supabase/client.py", line 61, in __init__
    options.headers.update(self._get_auth_headers())
AttributeError: 'dict' object has no attribute 'headers'

**Desktop (please complete the following information):**
 - OS: macOS

Simar-Kohli:
I figured a workaround: 
```
from supabase.lib.client_options import ClientOptions

clientOptions = ClientOptions(postgrest_client_timeout=999999)
supabaseClient: Client = create_client(SUPABASEURL, SUPABASEKEY, clientOptions)
```

Seems to solve the issue. Not sure if the website docs are out of date or maybe I'm still doing something wrong.

Simar-Kohli:
Well while this works on my local machine, this does not run on an EC2 instance. Would love some assistance

jtewright:
I experienced this problem too, and found that **in addition** to setting `postgrest_client_timeout` when initiating the python client, you have to change the default `statement_timeout` for the correct role through the Supabase dashboard SQL Editor, as per https://supabase.com/docs/guides/database/postgres/configuration#changing-the-default-timeout.

[...]

Issue #376, opened 2023-02-15
setting timeout parameter does not change timeout


mohnish7:
**Describe the bug**
Despite setting a `timeout` parameter in clientOptions, the query still times out in 5 seconds

**To Reproduce**
```
from supabase import create_client, Client
from supabase.lib.client_options import ClientOptions

client_options = ClientOptions(timeout=60)
supabase = create_client(SUPABASE_URL, SUPABASE_ANON_KEY, options=client_options)
```

**Expected behavior**
The function should now time out after 60 seconds

**Screenshots**
<img width="259" alt="Screen Shot 2023-02-15 at 2 04 18 PM" src="https://user-images.githubusercontent.com/7302380/219127526-32ac6069-2c8e-449e-aa5f-a1df883c4abf.png">

**Additional context**
I am running Supabase version 1.0.0

J0:
Hey @mohnish7,

The timeout param currently isn't being passed through - a [fix was made](https://github.com/supabase-community/supabase-py/pull/369) but hasn't been published. I'll release it sometime this weekend - feel free to bump if you need it sooner

dschnabel:
As per my [comment](https://github.com/supabase-community/supabase-py/issues/293#issuecomment-1429464174) the fix didn't seem to work for me. Does it work for others?

J0:
Hey @dschnabel,

Just seeing this- it’s currently being passed to the postgrest client which may be why the fix doesn’t work out.  I’ll see if  I can pass it into the Auth client as well.

mohnish7:
Hey @J0 were you able to successfully release this fix last weekend?

mohnish7:
Hey @J0 following up again!

J0:
Hey @mohnish7,

Thanks for the tag, so sorry didn't realize I missed this - wasn't able to get around to it but it's on my todo list for this week

rotemCyberillium:
Hey @J0 

I'm dealing with the the same problem..

enjalot:
I'm also dealing with this issue

jon-qualifai:
Hey @J0 Seeing the same issue with calling Supabase from Python and seeing timeouts. Any luck with this issue?

kernelguardian:
This might work
`
opts = ClientOptions().replace(postgrest_client_timeout=60)
supabase: Client = create_client(url, key, options=opts)
`

J0:
Hey team,
[...]

PR #773, opened 2024-04-17
Fix timeout

juancarlospaco:

Fix for timeout-related issues.

According to httpx documentation setting the timeout property to `None` will ensure no timeout is used.

This could be useful when the user cannot estimate how long a running task should take, 
and improves the user experience because avoids using arbitrary values to get rid of such timeout errors.

A test to play with timeouts is included, with happy-path and expected-to-fail scenarios.

See also:

- https://github.com/encode/httpx/blob/4b85e6c3898b94e686b427afd83138c87520b479/httpx/_client.py#L81
- https://github.com/supabase-community/postgrest-py/blob/4b3a664abf2c35b7a08919ae7bf71c90f3fa9ef8/postgrest/constants.py#L6

Explicit is better than implicit.

- https://github.com/supabase-community/supabase-py/issues/487
- https://github.com/supabase-community/supabase-py/issues/376
- https://github.com/supabase-community/supabase-py/issues/509#issuecomment-1721348841
- https://github.com/supabase-community/supabase-py/issues/510 ???

## Notes

Another different potential approach can be just increasing the `DEFAULT_POSTGREST_CLIENT_TIMEOUT`,
a value of `5` feels too small for long-running tasks whatsoever.
JavaScript client does not have an explicit timeout **by default** (?), AFAIU.
Some libraries in the wild use `-1` or `0` to mean no explicit timeout or a very huge number likely unreachable.
If the desired approach is to use an implicit integer, then maybe using a huge number can be considered.

### DIFF
--- a/supabase/lib/client_options.py
+++ b/supabase/lib/client_options.py
@@ -33,9 +33,7 @@ class ClientOptions:
     realtime: Optional[Dict[str, Any]] = None
     """Options passed to the realtime-py instance"""
 
-    postgrest_client_timeout: Union[
-        int, float, Timeout
-    ] = None
+    postgrest_client_timeout: Union[int, float, Timeout] = None
     """Timeout passed to the SyncPostgrestClient instance."""
 
     storage_client_timeout: Union[int, float, Timeout] = DEFAULT_STORAGE_CLIENT_TIMEOUT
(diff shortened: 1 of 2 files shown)


J0:
Hey! 

Personally, I think having `None` is fine. Thanks much for the tests :)

silentworks:
Changing this to `None` would be a breaking change. I'd prefer if the timeout was increased instead and we push for `None` on the next major release.

juancarlospaco:
See also: https://github.com/supabase-community/postgrest-py/pull/417#issue-2268986868

PR #846, opened 2024-07-03
Fix #814

### LINKED ISSUE #814: httpx.ReadTimeout: timed out Supabase functions
# Feature request

## Is your feature request related to a problem? Please describe.

When calling an Edge Function using `client.functions.invoke()`, if the function takes longer than 5 seconds to run I'll get a `ReadTimeout` exception

## Describe the solution you'd like

I would like a way to configure the HTTPX timeout for Edge Function calls, the same way we can currently configure PostgREST and Storage. Or a way to supply a custom HTTPX client to supafunc-py

## Describe alternatives you've considered

The following will let me create a client with a longer timeout, but it's accessing private interfaces:

```
supabase: Client = create_client(url, key, options=client_options)
func_headers = supabase.functions.headers
func_url = supabase.functions.url
timeout_client = utils.SyncClient(timeout=300.0, base_url=func_url, headers=func_headers, follow_redirects=True)
supabase.functions._client = timeout_client
```

juancarlospaco:

- Fix #814
- Requires https://github.com/supabase-community/functions-py/pull/120

### DIFF
--- a/supabase/_async/client.py
+++ b/supabase/_async/client.py
@@ -195,13 +195,6 @@ class AsyncClient:
             )
         return self._functions
 
-    def functionTimeout(self, timeout: int):
-        """Set function timeout"""
-        if self._functions:
-            assert timeout > 0, "Timeout must be positive."  # No negative timeout.
-            self._functions._client.timeout = min(
-                int(timeout), 150
-            )  # >150 returns HTTP-504.
 
     #     async def remove_subscription_helper(resolve):
     #         try:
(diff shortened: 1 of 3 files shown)


mansueli (approved):

silentworks (changes requested):
This should follow how we are setting `postgrest_client_timeout` in other libraries. It should be passed through the constructor and not be added via a method through a private variable.

J0 (approved):

silentworks (approved):

PR #917, opened 2024-09-02
fix: Cleanout old code

juancarlospaco:

- Cleanout future imports `from __future__ import annotations` no longer needed for Python >= `3.9.0`.

### DIFF
--- a/supabase/_async/__init__.py
+++ b/supabase/_async/__init__.py
@@ -1,1 +1,0 @@
-from __future__ import annotations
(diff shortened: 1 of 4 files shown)


grdsdev (approved):

juancarlospaco:
Closing this for now sorry, maybe in the future we can revisit the idea, 
the official status is kinda "TBD" even if in practice works for years now.

photo_2024-09-04_09-15-36

PR #920, opened 2024-09-03
Await for realtime set auth on auth event

LuisLechugaRuiz:

Solves https://github.com/supabase/supabase-py/issues/919.

Needs: https://github.com/supabase/auth-py/pull/591

AsyncClient crashes due to missing await at [self.realtime.set_auth()](https://github.com/supabase/supabase-py/blob/main/supabase/_async/client.py#L301)

## What is the new behavior?

AsyncClient now awaits for self.realtime.set_auth

### DIFF
--- a/supabase/_async/client.py
+++ b/supabase/_async/client.py
@@ -284,7 +284,7 @@ class AsyncClient:
             "Authorization": authorization,
         }
 
-    def _listen_to_auth_events(
+    async def _listen_to_auth_events(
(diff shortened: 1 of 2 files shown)


silentworks (changes requested):
This should be solved at the realtime client level and not in the function that you have changed here.

silentworks:
We've made a change to fix this in another PR and will release a new version soon. https://github.com/supabase/supabase-py/pull/930

PR #928, opened 2024-09-20
fix: update exports from init file

silentworks:

Bug fix to improve the use of the library.

Async client and errors aren't being exported from the `__init__.py` file.

## What is the new behavior?

Async client and errors are being exported from the `__init__.py` file.

## Additional context

### DIFF
--- a/supabase/__init__.py
+++ b/supabase/__init__.py
@@ -1,6 +1,18 @@
+from gotrue.errors import (
+    AuthApiError,
+    AuthError,
+    AuthImplicitGrantRedirectError,
+    AuthInvalidCredentialsError,
+    AuthRetryableError,
+    AuthSessionMissingError,
+    AuthUnknownError,
+    AuthWeakPasswordError,
+)
 from postgrest import APIError as PostgrestAPIError
 from postgrest import APIResponse as PostgrestAPIResponse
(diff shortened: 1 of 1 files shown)


supabase/__init__.py
```diff
@@ -1,35 +1,66 @@
...
+    AuthRetryableError,
+    AuthSessionMissingError,
+    AuthUnknownError,
+    AuthWeakPasswordError,
+)
 from postgrest import APIError as PostgrestAPIError
 from postgrest import APIResponse as PostgrestAPIResponse
+from realtime import AuthorizationError, NotConnectedError
 from storage3.utils import StorageException
+from supafunc.errors import FunctionsError, FunctionsHttpError, FunctionsRelayError
```
J0, line 15: Not urgent: I think there was an ask for a move from `supafunc` to `supabase_functions` can do that in the future

J0 (approved):

supabase/__init__.py
silentworks: Yes this has to wait until 3.0 of the library as it's a breaking change also this current addition is to help with migrating to that as user's will now import the errors from the `supabase` library rather than `supafunc` as some are probably doing now.

Issue #946, opened 2024-09-30
ERROR:root:join push timeout for channel realtime


blockfer-rp:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug
The Supabase real-time client is experiencing frequent disconnections and reconnection attempts. The client fails to maintain a stable connection, with repeated join push timeouts and keepalive ping timeouts.

## To Reproduce

1. Set up a Supabase project with real-time enabled
2. Create a Python script using the following code:

```python
import asyncio
import logging
from supabase._async.client import AsyncClient as Client, create_client

logging.basicConfig(level=logging.DEBUG)

supabase_url = "https://dmwfqahwcgkglvwyzhiy.supabase.co"
supabase_key = "your_supabase_key_here"

def handle_changes(payload):
    logging.info(f"Change received: {payload}")

async def test_realtime():
    supabase: Client = await create_client(supabase_url, supabase_key)
    
    try:
        await supabase.realtime.connect()
        logging.info("Connected to Supabase realtime")

        channel = supabase.channel('trade-tokens-updates')
        channel = channel.on_postgres_changes(
            event="INSERT",
            schema="public",
            table="trade_tokens",
            callback=handle_changes
        )
        
        await channel.subscribe()
        logging.info("Subscribed to channel")
        
        await asyncio.sleep(300)  # Wait for 5 minutes
    except Exception as e:
        logging.error(f"An error occurred: {e}")
    finally:
        if 'channel' in locals() and channel:
            await channel.unsubscribe()
        await supabase.realtime.disconnect()

if __name__ == "__main__":
    asyncio.run(test_realtime())
```
    
## Expected behavior
The client should maintain a stable connection to the Supabase real-time service, successfully subscribing to the channel and staying connected without frequent timeouts or disconnections.

## System information
supabase: 2.8.0
realtime: 2.0.5

Sanidhyafeaturist:
I will see what should be done

visheshdvivedi:
This issue might be related to the 'realtime' package and not 'supabase-py' package. In that case, we need to close this issue and add one issue there.

<img width="625" alt="image" src="https://github.com/user-attachments/assets/b2a80f40-2319-4335-beda-e580a815d3cd">

rdong8:
[...]

PR #975, opened 2024-10-22
fix: root:join push timeout for channel realtime #946 fix #974

Sanidhyafeaturist:


silentworks:
The reconnect logic you have here should be in the `realtime-py` library and not this library. Also you've added other changes to this PR which we would not be able to merge even if this PR was in the correct repo.

PR #977, opened 2024-10-29
feat: do not send non-JWTs in Authorization header

silentworks:

In preparation for the introduction of new API keys, the client library should not pass non-JWT values in the Authorization header, as requests like this will definitely fail.

## What is the new behavior?

## Additional context

https://github.com/supabase/supabase-js/pull/1293


MamlakatAlGhzalan:
I

PR #1013, opened 2024-12-13
chore: update httpx

ex0ns:

The last bump of httpx has been done in this commit: https://github.com/supabase/supabase-py/commit/fadbc4b41fc7ef2dc92929c28150901ea700f83f

httpx now release 0.28 (and 0.28.1) and this is causing some conflicts with other packages.

Unable to install supabase alongside some packages because of httpx

## What is the new behavior?

It should not change the existing install but not block project requiring httpx >= 0.28

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2115,4 +2115,4 @@ type = ["pytest-mypy"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9"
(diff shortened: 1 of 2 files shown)


ex0ns:
I ran `poetry lock --no-update` to avoid changes to the lockfile, but I'm not 100% confident that this is the proper way of doing this change.

silentworks (approved):

PR #1045, opened 2025-01-30
fix: update SupabaseAuthClient to use super instead of calling base class

PedroManuelAtienzaHuerta:
Hi, my name is Pedro Manuel. I have modified the file to use super() instead of calling the base class constructor directly. This makes the code clearer and easier to maintain.

File update.


juancarlospaco:
LGTM.

silentworks (approved):

Issue #1086, opened 2025-03-26
supabase-py self-hosted with self-signed certificate verify CLient Option not exposed


nibler-fox-kiki:
# Bug report

- [ x ] I confirm this is a bug with Supabase, not with my own application. 
- [ x ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Unable to access httpx `verify` kwarg from the `create_client` options. The `verify` kwargs are clearly available in _sync/client.py and _async/client.py but does not seem to be exposed to the `SyncClientOptions`/`AsyncClientOptions`

## To Reproduce

```python
supabase: Client = create_client(settings.SUPABASE_URL, settings.SUPABASE_KEY, options=SyncClientOptions(verify=False))

>>> TypeError: SyncClientOptions.__init__() got an unexpected keyword argument 'verify'
```

## Expected behavior

`verify` kwarg should be passed to the httpx client to allow working with self-signed certificate

silentworks:
This can now be done in version 2.16.0 using your own custom `httpx` client. 

```py
import certifi
import ssl
from httpx import Client
from supabase import create_client

# This SSL context is equivelent to the default `verify=True`.
ctx = ssl.create_default_context(cafile=certifi.where())

with Client(verify=ctx) as http_client:
    # Create a client with the custom httpx client
    options = ClientOptions(httpx_client=http_client)

    client = create_client(supabase_url, supabase_key, options)
```

nibler-fox-kiki:
With supabase-py 2.16.0 trying to use the below to pass `verify=False` result in the trace below:

```python
Traceback (most recent call last):
File "playground/supabase/connection_manager.py", line 27, in connect_supabase
    with Client(verify=ctx) as http_client:
         ~~~~~~^^^^^^^^^^^^
TypeError: SyncClient.__init__() got an unexpected keyword argument 'verify'
```


Example code:
```python
import os

import ssl
from httpx import Client
from dotenv import load_dotenv
from supabase import create_client, Client
from supabase.lib.client_options import ClientOptions

load_dotenv()

def connect_supabase():
    # This SSL context is equivalent to the default `verify=True`.
    # ctx = ssl.create_default_context(cafile=certifi.where())

    # Create an SSL context that does NOT verify certificates
    ctx = ssl.create_default_context()
    ctx.check_hostname = False
[...]

PR #1091, opened 2025-03-31
fix: expose verify option through client options

silentworks:

Bug fix to expose `verify` via the `ClientOptions` through the `httpx_options` config.

```py
from supabase import create_client, ClientOptions

supabase = create_client(
  url, 
  anon_key, 
  options=ClientOptions(httpx_options={"verify": False})
)
```

There is no way to set the `verify` flag from the `supabase` python library.

## What is the new behavior?

You can now set the `verify` flag from the `supabase` python library

## Additional context

#1086

### DIFF
--- a/supabase/_async/client.py
+++ b/supabase/_async/client.py
@@ -81,6 +81,11 @@ class AsyncClient:
         self.auth = self._init_supabase_auth_client(
             auth_url=self.auth_url,
             client_options=options,
+            verify=(
+                options.httpx_options.get("verify")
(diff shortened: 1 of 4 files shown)


grdsdev (approved):

silentworks:
Do not merge as yet, we are looking into a better option which won't need us replicating httpx options.

grdsdev (changes requested):
blocking until better option

silentworks:
Closing in favor of #1117

PR #1092, opened 2025-03-31
feat: expose httpx verify flag through ClientOptions

Orrv2904:
Resolves #1091

## Descripción
Permite configurar la verificación SSL mediante:
```python
create_client(url, key, options=SyncClientOptions(httpx_options={"verify": False}))


silentworks:
There is already a PR opened for this.

Issue #1095, opened 2025-04-02
Mutable reference options.headers


AsahiSoftWareEngineer:
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Mutable reference `options.headers`

`options.headers` is updated in `create_client` directry. that`s why, The headers may be changed unintentionally if it is shared `ClientOptions` instance other position.

## To Reproduce

Please execute below source code jupytor notebook. 
```python
from supabase import create_client
from supabase.lib.client_options import SyncClientOptions 
class settings:
    SUPABASE_URL = "https://***"
    SUPABASE_KEY = "***"
from supabase._sync.client import create_client, ClientOptions
from gotrue import SyncMemoryStorage

# Share same ClientOptions Instanc
shared_options = ClientOptions(storage=SyncMemoryStorage(), headers={"Custom-Header": "InitialValue"})

# Create client1
client1 = create_client(settings.SUPABASE_URL, settings.SUPABASE_KEY, options=shared_options)

# Create client2
client2 = create_client(settings.SUPABASE_URL, settings.SUPABASE_KEY, options=shared_options)

# Rewrite header client1
client1.options.headers["Custom-Header"] = "ModifiedValue"
client2.options.headers["Custom-Header"] = "ModifiedValue2"

# Check header client2
print(client1.options.headers["Custom-Header"])  # "ModifiedValue2" 
```

## Expected behavior

```python
...
print(client1.options.headers["Custom-Header"])
# -> "ModifiedValue" 
```

## Screenshots

<img width="1203" alt="Image" src="https://github.com/user-attachments/assets/d0282c3b-0ef8-4382-a981-a0a600ff8030" />

## System information

- mac OS
- chrome
- Version of supabase-py: 2.15.0
- Version of Python: 3.9.0

AsahiSoftWareEngineer:
@AsahiSoftWareEngineer I am going to fix bugs

gayathriramanesh:
Hi @AsahiSoftWareEngineer Is this issue fixed?

grdsdev:
Hey all, as discussed in https://github.com/supabase/supabase-py/pull/1096 this isn't a bug, it is just how Python works.

If you want to change a client's options you need to use the `replace` method in the option.

[...]

PR #1096, opened 2025-04-02
fix: mutable reference headers #1095

AsahiSoftWareEngineer:

Bug fix Isses #1095 

Isses No. #1095 

## What is the new behavior?

Prevent unintended header changes when sharing the same `ClientOptions`.


AsahiSoftWareEngineer:
@olirice Sorry, I can not set reviewer. That's why , I request a review here.
 Please review mu pull request.

supabase/_sync/client.py
```diff
@@ -293,8 +290,11 @@ def _listen_to_auth_events(
             self._storage = None
             self._functions = None
             access_token = session.access_token if session else self.supabase_key
-
-        self.options.headers["Authorization"] = self._create_auth_header(access_token)
+        header = copy.deepcopy(self._create_auth_header(access_token))
+        self.options.headers["Authorization"] = header
+        self.auth._headers["Authorization"] = header
+        self.postgrest.session.headers["Authorization"] = header
+        self.storage.session.headers["Authorization"] = header
```
grdsdev, line 298: not sure the issue this is solving, could you elaborate?

tests/test_client.py
```diff
@@ -113,9 +113,41 @@ def test_updates_the_authorization_header_on_auth_events() -> None:
...
+
+    authorization = "Bearer secretuserjwt"
+    options = ClientOptions(headers={"Authorization": authorization})
+
+    client = create_client(url, key, options)
+
+    assert client.options.headers.get("apiKey") == key
+
+
+# 実行
```
grdsdev, line 153: Remove comments
```suggestion
```

grdsdev (changes requested):
Thanks for the fix @AsahiSoftWareEngineer just some clarifications before merging.

supabase/_sync/client.py
AsahiSoftWareEngineer: Sorry, Below source code is not necessary.
```python
  self.postgrest.session.headers["Authorization"] = header
  self.storage.session.headers["Authorization"] = header
```

Caused by using deepcpy, before source code was not working. (auth._headers is not refreshed). 
That's why,  I changed to reassign deep-copied `self._create_auth_header(access_token)`methods

AsahiSoftWareEngineer:
@grdsdev Sorry, Please re-review pull request?

grdsdev (changes requested):

supabase/_async/client.py
```diff
@@ -69,7 +70,7 @@ def __init__(
 
         self.supabase_url = supabase_url
         self.supabase_key = supabase_key
-        self.options = options
+        self.options = copy.deepcopy(options)
         options.headers.update(self._get_auth_headers())
```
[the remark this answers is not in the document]
AsahiSoftWareEngineer: Rewrite `options.headers.update(self._get_auth_headers())` to `self.options.headers.update(self._get_auth_headers())`.

silentworks:
I think this would be better addressed by using the `.replace` method on the `ClientOptions` class. This should probably be added to the docs as the go to way of updating anything inside of the ClientOptions.

I wrote your test below using the `.replace` method and it all passes.

```py
def test_mutable_headers_issue():
    url = os.environ.get("SUPABASE_TEST_URL")
    key = os.environ.get("SUPABASE_TEST_KEY")

    shared_options = ClientOptions(
        headers={"Authorization": "Bearer initial-token", "x-site": "supanew.site"}
    )

    client1 = create_client(url, key, shared_options)
    client2 = create_client(url, key, shared_options)
    client1.options.replace({"headers": {"Authorization": "Bearer modified-token"}})

    assert client2.options.headers["Authorization"] == "Bearer initial-token"
    assert client2.options.headers["x-site"] == "supanew.site"
    assert client1.options.headers["x-site"] == "supanew.site"
```

AsahiSoftWareEngineer:
@silentworks Sure. I think of trying, but I do not know Can I rewrite docs.(Have I permissions)

AsahiSoftWareEngineer:
All test passed

supabase/_async/client.py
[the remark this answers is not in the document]
silentworks: It doesn't look like this change was made. Also note you only need to make changes in the `_async/client` and then run `make build_sync` and `make tests_pre_commit` in the terminal after.

supabase/_sync/auth_client.py
silentworks, line 1: Please revert this file as the type info in the doc blocks are incorrect. I think the `make build_sync` command is causing this issue and something we will fix. For now just revert this file.

AsahiSoftWareEngineer:
@grdsdev Hi, Can you re-review pull request.

silentworks:
@grdsdev this is fine. I do however think we are trying to fix a language (python being dynamic by design) issue here but I am OK with this PR being merged in.

grdsdev:
> @grdsdev this is fine. I do however think we are trying to fix a language (python being dynamic by design) issue here but I am OK with this PR being merged in.

that makes sense, let's not fight the language.

grdsdev:
@AsahiSoftWareEngineer thank you for this contributions, but as said by @silentworks there're other ways of achieving this by using the `replace` method.

silentworks:
We're re-opening this PR as I've realised in my tests I didn't test it fully.  `.replace` doesn't work as how I had initially thought as it returns a new object. In order to use `.replace` and get the same result as this PR the code would look more like the test below which I think might be too verbose for some.

```python
def test_mutable_headers_issue():
    url = os.environ.get("SUPABASE_TEST_URL")
    key = os.environ.get("SUPABASE_TEST_KEY")

    shared_options = ClientOptions(
        headers={"Authorization": "Bearer initial-token", "x-site": "supanew.site"}
    )

    client1 = create_client(url, key, shared_options)
    client2 = create_client(url, key, shared_options)
    client1.options = client1.options.replace(headers={**client1.options.headers, "Authorization": "Bearer modified-token"})

    assert client1.options.headers["Authorization"] == "Bearer modified-token"
    assert client2.options.headers["Authorization"] == "Bearer initial-token"
    assert client2.options.headers["x-site"] == "supanew.site"
    assert client1.options.headers["x-site"] == "supanew.site"
```

silentworks (approved):

grdsdev (approved):

PR #1151, opened 2025-06-20
fix: remove jwt key validation to allow new api keys

grdsdev:

Close https://github.com/supabase/supabase-py/issues/1150

## What is the new behavior?

## Additional context

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import os
 from unittest.mock import MagicMock
 
+from gotrue import SyncMemoryStorage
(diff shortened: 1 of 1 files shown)


silentworks (approved):

silentworks:
@grdsdev I'm now wondering if we should check if its not the new api key by checking for the `sb_` prefix first and then still do these JWT checks otherwise?

silentworks:
Actually ignore my last comment. Lets remove these check all together. I just checked the JS libraries and we aren't doing such checks since the auth server will handle the validation.

Issue #1166, opened 2025-07-03
Packages name change


silentworks:
# Chore

Migrate to the new package names and stop maintaining the older package names

## Describe the chore

We need to fully migrate from the `gotrue` package name to `supabase_auth`. Currently, we maintain support for both packages by using a CLI workaround that copies all files and applies a regex to replace `supabase_auth` with `gotrue` before publishing.

However, this approach has become problematic-most notably, it prevents us from installing the package locally for development and testing. I’d like to deprecate `gotrue` entirely and move forward exclusively with `supabase_auth`.

The same issue exists with the transition from `supafunc` to `supabase_functions`. We're still using the old package imports in the supabase-py package to avoid breaking backward compatibility for users who may be importing exceptions or errors directly from these packages.

## Additional context

Add any other context or screenshots that help clarify the task.

grdsdev:
We should do a final release for the `gotrue` and `supafunc` packages, and include the `classifiers` as:

```python
classifiers=[
    "Development Status :: 7 - Inactive",
],
```

Also, introduce a warning when initializing the client through the deprecated package, or when importing them (not sure if it's possible).

The warning message should state that the package name isn't supported anymore, and redirect it to the new package name.

```python
import warnings

warnings.warn(
    "The 'old-package-name' is deprecated. Use 'new-package-name' instead.",
    UserWarning,
    stacklevel=2
)
```

We can then post an announcement on discussions about it, also say that old package names aren't receiving new features, and only patches for X months.

wdyt @silentworks and @o-santi ?

o-santi:
I agree with that approach.

developerlee79:
Hi. Are the final release(deprecation) for the `gotrue` and `supafunc` packages, and the migration to the new package names,  open for contributions?

silentworks:
@grdsdev I don't think those changes alone will do it. There is a scenario where a user might be making use of the AuthAPIError from gotrue based on knowing `supabase-py` installs it. I think the first step is to make all the exception imports available directly from `supabase-py` first. 

So instead of a user importing from `gotrue.errors`:
```py
from gotrue.errors import (
    AuthApiError,
    AuthInvalidCredentialsError,
    AuthWeakPasswordError,
)
``` 
they would import from `supabase` or even `supabase.errors`:
```py
from supabase.errors import (
    AuthApiError,
    AuthInvalidCredentialsError,
    AuthWeakPasswordError,
)
```
[...]

PR #1190, opened 2025-08-04
chore: switch to monorepo, move realtime inside

o-santi:

This PR contains the first change in the monorepo switch, by moving `realtime` into this repository, and deprecating https://github.com/supabase/realtime-py (which I intend to do later). This requires restructuring this repository's files in order to leverage `uv` workspaces, as explained in https://github.com/supabase/supabase-py/pull/1186. 

The changes can be summarized by the following tree diagram:
```
.
├── CHANGELOG.md
├── CODE_OF_CONDUCT.md
├── CONTRIBUTING.md
├── LICENSE
├── MAINTAINERS.md
├── README.md
├── docs
│   └── ...
├── release-please-config.json
├── flake.lock                  <- mostly copied from realtime-py
├── flake.nix                   <- mostly copied from realtime-py
├── justfile                    <- new
├── pyproject.toml              <- virtual workspace root
├── src                         <- new packages directory
│   ├── realtime                <- inlined from supabase/realtime-py
│   │   ├── README.md           <- inlined from supabase/realtime-py
│   │   ├── infra               <- inlined from supabase/realtime-py
│   │   │   └── ...
│   │   ├── mod.just            <- new
│   │   ├── pyproject.toml      <- inlined from supabase/realtime-py
│   │   ├── src                 <- inlined from supabase/realtime-py
│   │   │   └── realtime        <- inlined from supabase/realtime-py
│   │   │       ├── __init__.py
│   │   │       └── ...
│   │   └── tests               <- inlined from supabase/realtime-py
│   │       └── ...
│   └── supabase                <- moved to sub package
│       ├── README.md           <- moved to sub package
│       ├── mod.just            <- new
│       ├── pyproject.toml      <- supabase-py library root
│       ├── src                 <- moved to sub package
│       │   └── supabase        <- moved to sub package
│       │       ├── __init__.py <- moved to sub package
│       │       └── ...
│       └── tests               <- moved to sub package
│           └── ...
├── test.ps1
└── uv.lock

```

### File structuring
`uv` broadly expects python "packages" to be either libraries or applications; and for workspaces, it expects them to be structured as N libraries, whose directory need to be specified in `uv.workspace.members`, and 1 application, located in `./src/{appname}`. This means that projects in the root of the repository are generally interpreted as describing the application, but given that this repository shouldn't have no application, I decided to move most of the roots `pyproject.toml` into `src/supabase`, and use it as a workspace library entry by listing it in `members`. `uv` generally calls these "virtual workspaces" but I couldn't find much documentation around them anywhere.

### Package specific dependencies and commands
`uv` already handles package specific dependencies by leveraging `uv --exact --package {package} {cmd} ...`, so no additional care is required besides calling the command correctly. In order to make running these more streamlined, and to share/compose them into bigger ones (for instance, to run the same commands locally and in CI), I think it makes sense to rely on a command runner. I outlined the following broad requirements in order to decide which tool to use:

1. All python packages resolution should be done by `uv`, including test/dev time dependencies (by using `dependency-groups`). This is to ensure that we are not specifying tests differently to run locally vs in CI.
2. Each package should be able to cleanly state its commands locally, and there should be a way to compose these from outside, without worrying with issues such as names clashing or environment leaks. A use case for this is that `realtime` requires certain docker instances to be up before running `pytest`, while `supabase` does not. On the flip side, `supabase` relies on `unasync-cli` to generate the `_sync`, with adhoc `sed` commands, while `realtime` does not use any of it - I don't think we want to keep this behavior in the future, but I do not intend to change it now. 
3. It should be easy to install, should not add any performance overhead over running the commands manually, should work out of the box with the local setup and should be supported by all major platforms we use. 

I decided to give [just](https://github.com/casey/just) a try. It can be summarized as a `make` successor, [that aims to be a command runner instead of a build system](https://github.com/casey/just/tree/master?tab=readme-ov-file#what-are-the-idiosyncrasies-of-make-that-just-avoids). By being a command runner, it does not suffer from issues that `Makefile`s suffer, like that targets may be interpreted as files, and thus need to be marked as PHONY sometimes, and has niceties such as aliases, readable errors, recipe arguments, invoking from anywhere (not only from the workspace root). 

`just` also allows for [structuring the files as namespaced modules](https://just.systems/man/en/modules1190.html), which helps to avoid hackier solutions to name collisions, like prepending every command with the package name. With this, each package can declare its commands for testing, linting and publishing inside `src/{packge}/mod.just`, that is then imported as a module inside the `justfile` in the root of the workspace. This allows us to define targets like `just test` by composing each of the packages' `test`, without any additional coordination.
```just
mod supabase "src/supabase/mod.just"
mod realtime "src/realtime/mod.just"

test: realtime::test && supabase::test
```

#### Alternatives
I believe that trying to use `Makefile`s to the same effect would result in a big spaghetti mess of commands, which is why I sought for a different tool. `just` was the best tool I found that was also minimalistic and not an entire build system, but I'd love some more suggestions for monorepo related tooling, specially those that generally fit those guidelines. For the record, I considered the following tools:
- [tox](https://tox.wiki/en/4.28.4/) / [nox](https://nox.thea.codes/en/stable/): requires dependencies to be installed by their bespoke system, which would require duplicating a lot of version requirements in different places
- [invoke](https://www.pyinvoke.org/): requires an outside python installation with invoke, which would require one to use `uv` to run it in the first place.
- [pants](https://www.pantsbuild.org/): bespoke entire build system, does not integrate with `uv`.


silentworks:
Great work and explanations here @o-santi. I really like the direction this is going but I think we need to ensure #1166 happens before this PR get's merged in. We would need to make at least two more releases with the current setup first before migration to this setup. Otherwise we will still be releasing two versions of the `auth` and `functions` packages.

o-santi:
Of course, I just wanted to get some comments and suggestions before moving forward with this. In talking with @olirice, we came to the conclusion that it might make sense to leave the command runner part out of the PR for now, and after all the sub packages are merged in, we can try to find the best way to do it.

o-santi:
Now that both packages have been deprecated, we can finally move on with this MR. I believe for this to get merged there are two remaining questions:
1. What are we going to do with the command runner? I do not think adding a new dependency for this is bad, but are we okay with `just`? I'd prefer it to `Makefile`s. We can not commit to any tool for now and change it later, but it will surely make the github CI workflow a mess for a while, and local testing won't be as easy.
2. More importantly, what about release versions? @mandarini is pushing for synchronized releases for all sub packages in the JS monorepo approach, in order to ensure that all packages are synchronized. I would prefer that approach too, but I'm not entirely sure what to do in the intermediary state where some packages have been migrated while other haven't. Maybe we can just ignore releases while this is happening?

@grdsdev @silentworks looping both of you in on this. What do you think?

grdsdev:
1. I'm still unsure about adding `just` dependency for that, but we can go with it if you think would be better.
2. I also think synchronized release versions is the best option here, and I don't see the issue the the intermediary state you mentioned, once this PR gets merged and released, both supabase and realtime will share the same version, the remaining ones will still be on their independent versions.

o-santi:
1. I will remove it and keep the old makefiles while we decide on what to do. I will try to muster up a stronger arguments supporting a new dependency for that while I experiment with the monorepo.
2. What I mean by intermediary state is that the versions aren't synchronized at all right now, some are even different major versions. For instance, realtime is currently at version `v2.7.0`, but supabase is at `v2.18.0`. What should the next release of realtime be? If we're going for synchronized, it should be at the next release of supabase-py itself, so most likely `v2.19`, but this would be a non-sequential jump. This is even worse for `supabase_functions`, which is currently at `0.10.1`, and upgrading it would not only be a major version change, but also completely skip version 1. What is the best way to handle this? Should we just ignore the fact that it is jumping, and warn that this skip is special in the `CHANGELOG`? @mandarini your input here would be useful too.

grdsdev:
I vote for synchronizing versions with supabase-py independent of current version, and make it stand out in CHANGELOG that although it is a major bump, there aren't significant changes, and the bump was just for aligning versions with main library.

mandarini:
I agree with @grdsdev here!

silentworks:
I agree with the version sync too. I think the packages that are pre 1.x are already stable and have feature parity with the JS libraries.

PR #1192, opened 2025-08-07
chore: switch to `supabase_auth` and `supabase_functions`

o-santi:

Switch to `supabase-auth` instead of `gotrue`, and `supabase-functions` instead of `supafunc`. Fixes https://github.com/supabase/supabase-py/issues/1166. 

Should be merge after both of the packages are properly deprecated.

Using `gotrue` and `supafunc` packages.

## What is the new behavior?

Using `supabase_auth` and `supabase_functions` packages.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,9 +21,9 @@ requires-python = ">=3.9"
 dependencies = [
   "postgrest == 1.1.1",
   "realtime == 2.7.0",
-  "gotrue == 2.12.3",
+  "supabase_auth == 2.12.3",
   "storage3 == 0.12.1",
-  "supafunc == 0.10.1",
+  "supabase_functions == 0.10.1",
   "httpx >=0.26,<0.29",
 ]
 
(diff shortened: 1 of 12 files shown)


silentworks (approved):

o-santi:
@silentworks I do think that its good that we're talking about warning people for that edge case, but overall I don't think that that specific case breaking means that the change is not backwards compatible, as it is abusing the fact that python transitive dependencies leak to the final project. 

The interface we're trying to maintain is the one declared in `supabase/__init__.py`, and we're explictly re-exporting `gotrue` definitions through that. If the user is correctly importing the definitions `from supabase`, then no error is thrown. Sadly, we cannot prohibit the user from importing `gotrue` even if its not declared as a package, but internal dependencies are certainly not part of the public interface.

In a similar vein, if a user is using `supabase`'s dev environment's `unasync-cli` without explicitly declaring it, and we decide to switch our tooling, then it certainly isn't a breaking change (if the client functions stay the same overall), as we do not promise at any point that `unasync-cli` is part of the public interface, and anyone relying on it is wrong in doing so.

silentworks (approved):

PR #1202, opened 2025-08-18
utils: add maybe_single + unit tests

KyleSDeveloper:
This PR introduces the `maybe_single` utility function to handle optional unwrapping of lists/tuples into a scalar when appropriate.  

- Added `supabase/_utils.py` with `maybe_single` implementation.  
- Added `tests/test_utils_maybe_single.py` with coverage for `None`, scalars, singleton lists, multi-item lists, and tuples.  

All new tests pass.


o-santi:
This function is not needed in the supabase utilities, and it doesn't make sense to offer it in our library. Closing as spam.

PR #1204, opened 2025-08-19
chore: include functions-py into monorepo

o-santi:

More functions-py into the monorepo. 

## Additional context

I believe the check list for moving both this project and the next ones should be roughly the following:
- [X] Add it as an entry to the workspace in `pyproject.toml`
- [X] Copy the relevant files to `src/{project_name}` (including `CHANGELOG.md`, `LICENSE` and `Makefile`!)
- [X] Use the package directly from workspace in `src/supabase`, instead of from pypi.
- [X] Add its tests to the main `Makefile`, and the directory itself to `make publish`.
- [X] For projects using `unasync-cli`, switch away from it.
- [X] Add it to main README.md


src/functions/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+name = "supabase_functions"
+version = "0.10.1" # {x-release-please-version}
+description = "Library for Supabase Functions"
+authors = [
+  { name = "Joel Lee", email = "joel@joellee.org" },
+  { name = "Andrew Smith", email = "a.smith@silentworks.co.uk" },
+]
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/supabase/functions-py"
```
silentworks, line 11: This should probably be updated to match the mono repo setup

Issue #1048, opened 2025-01-30
Storage client blows up on 400s


jezell:
# Bug report

- [ x] I confirm this is a bug with Supabase, not with my own application. 
- [xx ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

If any of the storage APIs returns a 400 error, the storage client doesn't properly handle the response and blows up trying to parse it as JSON.

## To Reproduce

Do something that would return a 400 from the storage api (often happens with uploads when given incorrect parameters for example).

```
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "supabase_storage_provider.py", line 39, in aclose
    await (await self.provider.client).storage.from_(self.provider.bucket_name).upload(
  File "venv/lib/python3.12/site-packages/storage3/_async/file_api.py", line 520, in upload
    return await self._upload_or_update("POST", path, file, file_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/storage3/_async/file_api.py", line 493, in _upload_or_update
    response = await self._request(
               ^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/storage3/_async/file_api.py", line 54, in _request
    raise StorageApiError(resp["message"], resp["error"], resp["statusCode"])
                          ^^^^
  NameError: name 'resp' is not defined
```

## Expected behavior

expected result is an exception with a description of the problem, not blowing up on parsing JSON

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

silentworks:
Please provide reproducible steps. Saying `Do something that would return a 400 from the storage api ` is not reproducible steps.

juancarlospaco:
Can not reproduce.

P3st83:
Hi! I'd love to work on this issue. Can I take it?

[...]

PR #1227, opened 2025-09-24
fix: 400 error blow ups when reponse not json

sofc-T:

https://github.com/supabase/supabase-py/issues/1048 issue bug fix

very small change
I found the error to be failure to parse jsons. and added a safe way to parse json

Bug fix, feature, docs update, ...
Very small change to parse the json safely.

#1048 

## What is the new behavior?

## Additional context


o-santi:
Not the correct solution. The way to fix this is to use pydantic parsing, which should properly handle when the message is incorrect.

PR #1229, opened 2025-09-25
feat(realtime): add presence enabled flag on join payload

grdsdev:
## Summary
- Add `enabled` field to `RealtimeChannelPresenceConfig` type definition
- Add `_has_callback_attached` property to `AsyncRealtimePresence` class to detect if any presence callbacks are registered
- Modify channel subscription logic to automatically set `presence.enabled=true` when presence callbacks are attached or manually configured
- Add automatic resubscription functionality when presence callbacks are added to an already joined channel

## Test plan
- [x] Add unit tests for `_has_callback_attached` property
- [x] Add tests for presence config type validation with `enabled` field
- [x] Add integration tests for automatic `presence.enabled` flag setting
- [x] Add tests for automatic resubscription behavior when callbacks are added post-join
- [x] Verify all existing presence functionality continues to work as expected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/realtime/src/realtime/_async/channel.py
+++ b/src/realtime/src/realtime/_async/channel.py
@@ -191,12 +191,16 @@ class AsyncRealtimeChannel:
         else:
             config: RealtimeChannelConfig = self.params["config"]
             broadcast = config.get("broadcast")
-            presence = config.get("presence") or RealtimeChannelPresenceConfig(key="",enabled=False)
+            presence = config.get("presence") or RealtimeChannelPresenceConfig(
+                key="", enabled=False
+            )
             private = config.get("private", False)
 
-            presence_enabled = self.presence._has_callback_attached or presence.get("enabled", False)
+            presence_enabled = self.presence._has_callback_attached or presence.get(
+                "enabled", False
+            )
             presence["enabled"] = presence_enabled
-            
+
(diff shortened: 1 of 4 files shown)


src/realtime/src/realtime/_async/channel.py
```diff
@@ -453,6 +473,11 @@ def on_presence_leave(
         :return: The Channel instance for method chaining.
         """
         self.presence.on_leave(callback)
+        if self.is_joined:
+            logger.info(
+                f"channel {self.topic} resubscribe due to change in presence callbacks on joined channel"
+            )
+            asyncio.create_task(self._resubscribe())
```
grdsdev, line 480: @o-santi not sure if this `create_task` is appropriate here, should I hold a reference for the task to make sure it is finished/cancelled?

o-santi: Yes, `create_task` spawns it in the background, which is a problem because we won't know when its done. However, I'm not sure I understand the goal here. `on_presence_leave` is the time of registering the callback. Most of the time, the user won't even be subscribed when this is called. Did you mean to register a callback on the presence leave event for resubscribing?

o-santi (approved):

PR #1232, opened 2025-09-27
docs: add quick usage example to README and fix typo

be991:
This PR adds a quick usage example in the README so that new users can immediately see how to use supabase-py. and was also able to fix some typo in the README file


o-santi:
The `supabase` example is already given at `src/supabase/README.md`. The monorepo root README is meant to give instructions on how to use the tools of the monorepo itself.

PR #1235, opened 2025-09-30
feat(realtime): add support for broadcast replay configuration

grdsdev:
## Summary
- Added `ReplayOption` type for configuring broadcast replay
- Added `replay` field to `RealtimeChannelBroadcastConfig` 
- Added `BroadcastMeta` type for replay metadata
- Added `meta` field to `BroadcastPayload` to receive replay information

## Changes
This PR implements broadcast replay support following the pattern from realtime-js PR #540:

1. **ReplayOption**: New TypedDict with `since` (int) and optional `limit` (int)
2. **RealtimeChannelBroadcastConfig**: Added optional `replay` field
3. **BroadcastMeta**: New TypedDict with `replayed` (bool) and `id` (str)
4. **BroadcastPayload**: Added optional `meta` field

## Usage
```python
# Configure broadcast with replay
channel = client.channel('my-channel', {
    'config': {
        'broadcast': {
            'self': True,
            'ack': True,
            'replay': {
                'since': 1234567890,
                'limit': 100
            }
        }
    }
})

# Broadcast callback receives meta field
def on_broadcast(payload):
    if payload.get('meta', {}).get('replayed'):
        print(f"Replayed message: {payload['meta']['id']}")
```

## Test plan
- [ ] Verify type checking passes
- [ ] Test broadcast replay configuration
- [ ] Verify meta field is properly typed in callbacks

🤖 Generated with [Claude Code](https://claude.com/claude-code)


edgurgel (approved):

grdsdev:
> I don't think these changes in particular are enough to make it work, because `RealtimeChannelBroadcastConfig` is not directly sent over wire, but instead read from inside `subscribe`. I believe you need to read that information inside the function to generate the `config_payload` correctly.

@o-santi just wrote a few tests to confirm this, and the config is properly sent over, no need for additional changes, can you review it again?

o-santi (approved):
I see what is happening. `replay` is a subfield of `broadcast`, which is already read from inside subscribe, so it is sent normally by wire. Small nitpick on temp file but should be correct.

mandarini (approved):

PR #1245, opened 2025-10-04
fix: prevent httpx.Client base_url mutation across services

### LINKED ISSUE #1244: Shared httpx.Client causes base_url clobber between PostgREST and Storage - results in 404 and KeyError
## Describe the bug

When passing a custom `httpx.Client` via `SyncClientOptions(httpx_client=...)`, the same client instance is shared across all Supabase services (PostgREST, Storage, Auth, Functions). Each service mutates the `httpx.Client.base_url` when initialized, causing subsequent requests to other services to hit the wrong API endpoints.

**Symptoms:**
- **404 errors** when Storage methods hit `/rest/v1/object/list/...` instead of `/storage/v1/object/list/...`
- **Cryptic `KeyError: 'error'`** because the 404 response format differs between PostgREST and Storage APIs
- **Non-deterministic failures** depending on which service was accessed most recently

**Root Cause:**

The issue is in how `storage3` and `postgrest` clients handle injected `httpx_client`:

**storage3/_sync/client.py:77-80**
```python
if http_client is not None:
    http_client.base_url = base_url  # Mutates shared client!
    http_client.headers.update({**headers})
    return http_client
```

**Timeline:**
1. Storage service initialized → `httpx_client.base_url = ".../storage/v1/"` (Works)
2. PostgREST service accessed → `httpx_client.base_url = ".../rest/v1/"` **Clobbers storage URL**
3. Storage methods called → still points to `/rest/v1/` → 404 → `KeyError: 'error'`

## To Reproduce

```python
import httpx
from supabase import Client, create_client
from supabase.lib.client_options import SyncClientOptions

url = "https://your-project.supabase.co"
key = "your-anon-key"

# Create client with shared httpx instance
timeout = httpx.Timeout(10.0, read=60.0)
httpx_client = httpx.Client(timeout=timeout)
options = SyncClientOptions(httpx_client=httpx_client)
client: Client = create_client(url, key, options=options)

bucket = client.storage.from_("your-bucket")
print(f"Initial base_url: {bucket._client.base_url}")
# Output: https://your-project.supabase.co/storage/v1/

# First storage call works
bucket.list("some-path/")  # Works

# Touch PostgREST (causes base_url mutation)
_ = client.postgrest
print(f"After postgrest: {bucket._client.base_url}")
# Output: https://your-project.supabase.co/rest/v1/  WRONG!

# Second storage call fails
bucket.list("some-path/")  # KeyError: 'error'
```

## Expected behavior

Storage requests should always hit `/storage/v1/` endpoints regardless of whether PostgREST or other services have been accessed. The second `bucket.list()` call should succeed with the same results as the first call.

**Actual behavior:** After accessing PostgREST, Storage requests hit `/rest/v1/` endpoints, causing 404 errors and `KeyError: 'error'` exceptions.

## Screenshots

Terminal output demonstrating the bug:

```python
# Demonstrating base_url mutation
bucket = client.storage.from_("snapshots")
print(f"Initial base_url: {bucket._client.base_url}")
# Output: https://your-project.supabase.co/storage/v1/

bucket.list("some-path/")  # Works

_ = client.postgrest  # Touch PostgREST
print(f"After accessing postgrest: {bucket._client.base_url}")
# Output: https://your-project.supabase.co/rest/v1/  CLOBBERED!

bucket.list("some-path/")  # Fails with KeyError: 'error'
```

Error output:
```
KeyError: 'error'
  File "storage3/_sync/file_api.py", line 53, in _request
    response.raise_for_status()
  File "httpx/_models.py", line 829, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '404 Not Found' for url 'https://your-project.supabase.co/rest/v1/object/list/snapshots'

During handling of the above exception, another exception occurred:
  File "storage3/_sync/file_api.py", line 56, in _request
    raise StorageApiError(resp["message"], resp["error"], resp["statusCode"])
KeyError: 'error'
```

## System information

- **OS**: macOS 14.5 / Linux (Ubuntu 22.04)
- **Version of supabase-py**: 2.x (latest)
- **Version of storage3**: 0.x (latest)
- **Version of Python**: 3.12.9
- **httpx version**: (latest)

grdsdev:
## Summary
Fixes #1244 by preventing `httpx.Client` base_url mutation when a custom httpx client is shared across services.

## Problem
When passing a custom `httpx.Client` via `SyncClientOptions(httpx_client=...)`, the same client instance was shared across all Supabase services (PostgREST, Storage, Auth, Functions). Each service mutated the `httpx.Client.base_url` when initialized, causing subsequent requests to other services to hit the wrong API endpoints.

**Symptoms:**
- 404 errors when Storage methods hit `/rest/v1/object/list/...` instead of `/storage/v1/object/list/...`
- Cryptic `KeyError: 'error'` because the 404 response format differs between PostgREST and Storage APIs
- Non-deterministic failures depending on which service was accessed most recently

## Root Cause
The issue occurred in how services handle injected `httpx_client`:
```python
# storage3/_sync/client.py
if http_client is not None:
    http_client.base_url = base_url  # Mutates shared client!
```

**Timeline:**
1. Storage service initialized → `httpx_client.base_url = ".../storage/v1/"` (Works)
2. PostgREST service accessed → `httpx_client.base_url = ".../rest/v1/"` **Clobbers storage URL**
3. Storage methods called → still points to `/rest/v1/` → 404 → `KeyError: 'error'`

## Solution
Create independent copies of the `httpx_client` for each service using `copy.copy()`. This ensures each service gets its own instance with a separate `base_url` that can be mutated independently, while still sharing the underlying connection pool and transport configuration.

## Changes
- Modified `_sync/client.py` to copy httpx_client for auth, postgrest, storage, and functions services
- Modified `_async/client.py` with the same fix for async clients
- Added `test_httpx_client_base_url_isolation` for both sync and async clients
- Tests verify that accessing PostgREST after Storage no longer mutates Storage's base_url

## Test Plan
✅ New tests added: `test_httpx_client_base_url_isolation` (sync and async)
✅ All existing tests pass (34 passed, 12 xfailed, 86 xpassed)
✅ Fix verified to resolve the issue described in #1244

## Additional Notes
- The fix is minimal and uses Python's standard `copy.copy()` method
- Preserves the shared connection pool benefits of httpx
- Prevents base_url mutation conflicts between services
- Works for both sync and async clients

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/supabase/src/supabase/_async/client.py
+++ b/src/supabase/src/supabase/_async/client.py
@@ -77,9 +77,15 @@ class AsyncClient:
         self.functions_url = f"{supabase_url}/functions/v1"
 
         # Instantiate clients.
+        # Create a copy of client_options with a copied httpx_client for auth
+        # to prevent base_url mutation across services
+        auth_options = copy.copy(self.options)
+        if self.options.httpx_client is not None:
+            auth_options.httpx_client = copy.copy(self.options.httpx_client)
+
         self.auth = self._init_supabase_auth_client(
             auth_url=self.auth_url,
-            client_options=self.options,
+            client_options=auth_options,
         )
         self.realtime = self._init_realtime_client(
             realtime_url=self.realtime_url,
@@ -170,12 +176,18 @@ class AsyncClient:
     @property
     def postgrest(self):
         if self._postgrest is None:
+            # Create a copy of httpx_client to prevent base_url mutation across services
+            http_client = (
+                copy.copy(self.options.httpx_client)
+                if self.options.httpx_client is not None
+                else None
+            )
             self._postgrest = self._init_postgrest_client(
                 rest_url=self.rest_url,
                 headers=self.options.headers,
                 schema=self.options.schema,
                 timeout=self.options.postgrest_client_timeout,
-                http_client=self.options.httpx_client,
+                http_client=http_client,
             )
 
         return self._postgrest
@@ -183,17 +195,29 @@ class AsyncClient:
     @property
     def storage(self):
         if self._storage is None:
+            # Create a copy of httpx_client to prevent base_url mutation across services
+            http_client = (
+                copy.copy(self.options.httpx_client)
+                if self.options.httpx_client is not None
+                else None
+            )
             self._storage = self._init_storage_client(
                 storage_url=self.storage_url,
                 headers=self.options.headers,
                 storage_client_timeout=self.options.storage_client_timeout,
-                http_client=self.options.httpx_client,
+                http_client=http_client,
             )
         return self._storage
 
     @property
     def functions(self):
         if self._functions is None:
+            # Create a copy of httpx_client to prevent base_url mutation across services
+            http_client = (
+                copy.copy(self.options.httpx_client)
+                if self.options.httpx_client is not None
+                else None
+            )
             self._functions = AsyncFunctionsClient(
                 url=self.functions_url,
                 headers=self.options.headers,
@@ -202,7 +226,7 @@ class AsyncClient:
                     if self.options.httpx_client is None
                     else None
                 ),
-                http_client=self.options.httpx_client,
+                http_client=http_client,
             )
         return self._functions
 
--- a/src/supabase/src/supabase/_sync/client.py
+++ b/src/supabase/src/supabase/_sync/client.py
@@ -76,9 +76,15 @@ class SyncClient:
         self.functions_url = f"{supabase_url}/functions/v1"
 
         # Instantiate clients.
+        # Create a copy of client_options with a copied httpx_client for auth
+        # to prevent base_url mutation across services
+        auth_options = copy.copy(self.options)
+        if self.options.httpx_client is not None:
+            auth_options.httpx_client = copy.copy(self.options.httpx_client)
+
         self.auth = self._init_supabase_auth_client(
             auth_url=self.auth_url,
-            client_options=self.options,
+            client_options=auth_options,
         )
         self.realtime = self._init_realtime_client(
             realtime_url=self.realtime_url,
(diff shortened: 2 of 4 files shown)


o-santi:
I'm not convinced that this is the correct approach to fixing it. `copy.copy` is only a shallow copy, so any inner objects from `httpx_client` will still be shared. If we wanted to be absolutely sure that these are different objects, we'd need to do a `copy.deepcopy`, which not only is very slow but may be forgotten in some place and cause the error again. 

I think a much better approach is to avoid modifying input objects in `storage` and `postgrest`. I will focus on that instead and report back with the results.

o-santi (changes requested):
I don't think `copy.copy` is enough to fixing. At least, we need `copy.deepcopy` but even then, I'm not sure this is the correct approach. See my other comment for further info.

grdsdev:
@o-santi feel free to ditch this PR in favor of yours.

silentworks:
My thoughts on this is that we want to share the configuration like `headers` across requests but we don't want the `base_url` to be mutated across request (per service library use). For this instance I think it would be best to supply the `base_url` as part of the actual request only at `request` time.

I think in almost all of the libraries we have a `_request` method which then calls the underlying httpx `request` method with the `method, url, json=json` properties. I think at this point we could attach the `base_url` and it would then prevent it's mutation on the `httpx_client` object itself.

Below is an example of the code inside of `storage-py` that would get modified.

```py
async def _request(
    self,
    method: RequestMethod,
    url: str,
    json: Optional[dict[Any, Any]] = None,
) -> Response:
    try:
        response = await self._client.request(method, f"{self.base_url}{url}", json=json)
        response.raise_for_status()
        resp = exc.response.json()
        raise StorageApiError(resp["message"], resp["error"], resp["statusCode"])

    return response
```

What are your thoughts on this approach @o-santi?

o-santi:
Yep, I was basically going to do exactly that, minus the manual concatenation of the url. I will open a new PR with these changes shortly.

grdsdev:
@o-santi I'm not sure of the Python side, but if possible lets try to not do URL manipulation using raw string, and leverage some URL type/URL library to handle it.

silentworks:
Closing this out in preference for #1249

PR #1249, opened 2025-10-07
fix: do not mutate httpx client inside storage, postgrest and functions

o-santi:

Fix both `storage` and `postgrest` to not mutate external httpx client, given it is shared, by instead keeping their own copy of their parameters to pass into the request call. In order to properly manipulate URLs without mutation, I've decided to add the `yarl` dependency, which returns a new object upon path manipulation.

For the `postgrest` library, I've also decided to factor out all of the parameters that go into a request in a `RequestConfig` object, for it is very easy to miss a spot otherwise. I've had to change the tests to account for this. However, I think this greatly simplifies most of the libraries, as 99% of the code does not need to be async/sync, and can be stated only once in a further rewrite.

## Additional context

Should fix #1245.

### DIFF
--- a/src/supabase/src/supabase/_sync/auth_client.py
+++ b/src/supabase/src/supabase/_sync/auth_client.py
@@ -1,10 +1,10 @@
 from typing import Dict, Optional
 
 from supabase_auth import (
-    AuthFlowType,
     SyncGoTrueClient,
     SyncMemoryStorage,
     SyncSupportedStorage,
+    AuthFlowType,
 )
 from supabase_auth.http_clients import SyncClient
 
--- a/src/supabase/src/supabase/_sync/client.py
+++ b/src/supabase/src/supabase/_sync/client.py
@@ -10,7 +10,7 @@ from postgrest import (
 )
 from postgrest.constants import DEFAULT_POSTGREST_CLIENT_TIMEOUT
 from postgrest.types import CountMethod
-from realtime import RealtimeChannelOptions, SyncRealtimeChannel, SyncRealtimeClient
+from realtime import SyncRealtimeChannel, SyncRealtimeClient, RealtimeChannelOptions
 from storage3 import SyncStorageClient
 from storage3.constants import DEFAULT_TIMEOUT as DEFAULT_STORAGE_CLIENT_TIMEOUT
 from supabase_functions import SyncFunctionsClient
@@ -72,7 +72,7 @@ class SyncClient:
         self.rest_url = f"{supabase_url}/rest/v1"
         self.realtime_url = f"{supabase_url}/realtime/v1".replace("http", "ws")
         self.auth_url = f"{supabase_url}/auth/v1"
-        self.storage_url = f"{supabase_url}/storage/v1"
+        self.storage_url = f"{supabase_url}/storage/v1/"
         self.functions_url = f"{supabase_url}/functions/v1"
 
         # Instantiate clients.
--- a/src/supabase/tests/_sync/test_client.py
+++ b/src/supabase/tests/_sync/test_client.py
@@ -1,6 +1,6 @@
 import os
 from typing import Any
-from unittest.mock import MagicMock, Mock
+from unittest.mock import Mock, MagicMock
 
 import pytest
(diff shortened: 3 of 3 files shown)


o-santi:
I do think letting the `httpx` client be passed in was a mistake in general, for it adds way too much ambiguity on how parameters are merged. For instance, `base_url` should be entirely ignored if set in `httpx`, while the other parameters should be additive in general.

grdsdev (approved):

src/postgrest/tests/_async/test_client.py
```diff
@@ -35,6 +35,7 @@ def test_simple(self, postgrest_client: AsyncPostgrestClient):
                 "Content-Profile": "public",
             }
         )
+        print(session.headers)
```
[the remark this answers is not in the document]
o-santi: Indeed.

src/postgrest/src/postgrest/_async/client.py
```diff
@@ -5,6 +5,8 @@
 
 from deprecation import deprecated
 from httpx import AsyncClient, Headers, QueryParams, Timeout
+from httpx._types import HeaderTypes
```
silentworks, line 8: Unused import

src/postgrest/src/postgrest/_sync/client.py
```diff
@@ -5,6 +5,8 @@
 
 from deprecation import deprecated
 from httpx import Client, Headers, QueryParams, Timeout
+from httpx._types import HeaderTypes
```
silentworks, line 8: Unused import

src/storage/src/storage3/_async/bucket.py
```diff
@@ -3,6 +3,8 @@
 from typing import Any, Optional
 
 from httpx import AsyncClient, HTTPStatusError, Response
+from httpx._types import HeaderTypes
```
silentworks, line 6: I'm not keen on a import from private module here. But it's not a deal breaker.

silentworks:
> I do think letting the `httpx` client be passed in was a mistake in general, for it adds way too much ambiguity on how parameters are merged. For instance, `base_url` should be entirely ignored if set in `httpx`, while the other parameters should be additive in general.

It was either this or passing every parameter that `httpx` expects as a config parameter which felt like duplication work. Let's discuss in a new issue if you have another approach you think would work better.

src/storage/src/storage3/_async/bucket.py
o-santi: This is to support both `dict[str, str]` and `Headers` types.

o-santi:
> I will pull this code down and test against my local codebase.

Thank you, please let me know of the results.

> It was either this or passing every parameter that `httpx` expects as a config parameter which felt like duplication work.

The problem is that it is still happening, `headers` and `base_url` are such examples of arguments that can be set on both, and are ambiguous as to what setting both means. Furthermore, `RequestConfig` is basically a configuration class for `httpx`, which we might decide to just use directly instead. Setting things on it is much easier as we can assign immutable semantics, instead of httpx's mutable API, which is horrible for sharing configuration. I believe the only way forward is to expand something like `RequestConfig` into all libraries, and pass it around with immutable semantics, ditching the `httpx` client argument.

o-santi:
Just noticed that this problem is not only on these two libraries, `functions` also has it. `realtime` does not accept an `httpx` client, while `auth` accepts it but does store its own `url` and headers.

silentworks:
@o-santi  I'm getting an error while using these changes in my test setup locally. The error is below.
```
Request URL is missing an 'http://' or 'https://' protocol.
```

silentworks:
`Async/SyncBucketActionsMixin` class is missing the `base_url` and only has the `path` as the URL hence the error above. This needs to be updated to match the `Async/SyncStorageBucketAPI` use of `base_url`.

silentworks (changes requested):
`Async/SyncBucketActionsMixin` class is missing the `base_url` and only has the `path` as the URL hence the error above. This needs to be updated to match the `Async/SyncStorageBucketAPI` use of `base_url`.

PR #1255, opened 2025-10-09
chore(ci): include python 3.14 into test suite

o-santi:

Add python 3.14 into test suite in CI.

## Additional context

We should define a deadline to remove 3.9 from the supported versions. October 2025 (this month) is officially last month of security support for 3.9 ([per official docs](https://devguide.python.org/versions/)), so we need to decide on a date for deprecation.


o-santi:
It seems that the CI is also hitting https://github.com/supabase/supabase-py/pull/1252, while I wasn't able to do it manually -- most likely my environment was using the wrong version of python. We'll need to wait for the upstream fix of pydantic to get merged and published before using 3.14. 

This PR can also serve the purpose of ensuring that this library is compatible with the new version release.

PR #1268, opened 2025-10-24
chore(ci): change conventional commits to the same job ran in swift

o-santi:

The current CI conventional commits checker is extremely buggy, and fails randomly when markdown is included in the PR description. Instead, I substituted it by the swift one, which only checks the PR title.


grdsdev:
Please also remove the `.github/workflows/conventional-commits-lint.js` script.

o-santi:
Thank you for noticing.

grdsdev (approved):

PR #1270, opened 2025-10-27
the typo "sunbase-py"

BinaryFighter-01:
src/supabase/CHANGELOG.md — the typo "sunbase-py" appears on line 681. Replace that single line with the corrected text "supabase-py"


BinaryFighter-01:
Hi @grdsdev, I’ve fixed the typo as suggested. Could you please review this PR when you have time? Thanks!

o-santi:
The changelog is an append only historical document, so I don't think we need to correct this typo.

PR #1288, opened 2025-10-30
fix(auth): return auth_response from exchange_code_for_session instead of response dict

timu-jesse-ezell:

Bug fix

Method started returning dict instead of pydantic object

## What is the new behavior?

Return the pydantic object 

## Additional context

There seems have been a behavior change introduced recently, not sure this was intentional, this method used to return the pydantic model, but here it is returning the json. This is breaking our code after updating. Looks like other methods are still returning the pydantic model, so I think this was a mistake.

### DIFF
--- a/src/auth/src/supabase_auth/_async/gotrue_admin_api.py
+++ b/src/auth/src/supabase_auth/_async/gotrue_admin_api.py
@@ -6,10 +6,10 @@ from httpx import QueryParams, Response
 from pydantic import TypeAdapter
 
 from ..helpers import (
-    validate_uuid,
     model_validate,
     parse_link_response,
     parse_user_response,
+    validate_uuid,
 )
 from ..http_clients import AsyncClient
 from ..types import (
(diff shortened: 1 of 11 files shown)


jezell:
Previous implementation for reference, xform=parse_auth_response causes response to be a pydantic model, new code seemed to rely on manually parsing the model, but then returns the wrong one of the two objects.

```python


    async def exchange_code_for_session(self, params: CodeExchangeParams):
        code_verifier = params.get("code_verifier") or await self._storage.get_item(
            f"{self._storage_key}-code-verifier"
        )
        response = await self._request(
            "POST",
            "token",
            query={"grant_type": "pkce"},
            body={
                "auth_code": params.get("auth_code"),
                "code_verifier": code_verifier,
            },
            redirect_to=params.get("redirect_to"),
            xform=parse_auth_response,
        )
        await self._storage.remove_item(f"{self._storage_key}-code-verifier")
        if response.session:
            await self._save_session(response.session)
            self._notify_all_subscribers("SIGNED_IN", response.session)
        return response
     
```

o-santi:
> There seems have been a behavior change introduced recently, not sure this was intentional, this method used to return the pydantic model, but here it is returning the json. 

Yes, this was a mistake where the variable was only renamed in some places and not others. It wasn't caught because the function does not declare a return type, and `mypy` infers `Any` for functions with no declared return type, instead of the usual `None` (https://github.com/python/mypy/issues/9413).

o-santi:
Fixed in v2.23.0.

PR #1327, opened 2025-12-09
fix(storage): add query parameters option to download

o-santi:

Adds a query parameters argument to `download` method of storage client. This is done because the docs on storage advice the user to bypass caching by setting a unique query parameter on the url ([relevant link](https://supabase.com/docs/guides/storage/cdn/smart-cdn#bypassing-cache)). However, since we started using the URL builders, the client has been (correctly) ignoring any passed in query parameters for that function, so that was not possible to do.


o-santi:
> Approved, do we have tests for this?

Just merged two tests to ensure this behavior is maintained going forward.

Issue #1333, opened 2025-12-15
Storage uses deprecated `pydantic.config.Extra`


breid1313:
When using the storage library, there are a lot of warnings from pydantic about the use of deprecated `pydantic.config.Extra` noting that will be removed in pydantic 3.0. The preferred method is to use string literals.

sample warning message:
```
PydanticDeprecatedSince20: `pydantic.config.Extra` is deprecated, use literal values instead (e.g. `extra='allow'`). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.12/migration/
    class VectorBucket(BaseModel, extra=Extra.ignore):
```

documentation: [https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra)

o-santi:
Fixed by https://github.com/supabase/supabase-py/pull/1334

PR #1334, opened 2025-12-15
fix(storage): replace deprecated pydantic Extra with literal values

breid1313:
Replace `extra=Extra.ignore` with `extra="ignore"` to resolve Pydantic V2 deprecation warnings. The Extra enum is deprecated and will be removed in Pydantic V3.

Updates use of deprecated pydantic object in favor of string literals as their documentation suggests

#1333 

## What is the new behavior?

The behavior is the same, but the loud warning messages will no longer occur

## Additional context

n/a


breid1313:
@olirice / @silentworks small syntax change here for consideration

PR #1337, opened 2025-12-16
fix(realtime): prevent channel from closing during auto-reconnect

Rohitth10e:

Bug fix

After an unexpected WebSocket disconnect, the Realtime client reconnects but channels can remain stuck in CLOSED.
This happens because phx_leave and phx_join are sent back-to-back during reconnect, and the leave reply closes the channel before the join completes.

issue: [https://github.com/supabase/supabase-py/issues/1311]

## What is the new behavior?

On reconnect, the client skips sending phx_leave and directly rejoins the channel.
Channels now reliably return to JOINED state and subscriptions no longer time out after reconnect.

## Additional context

Verified locally against a real Supabase project by forcing a WebSocket disconnect and observing that the channel remains JOINED after auto-reconnect.

### DIFF
--- a/src/realtime/src/realtime/_async/channel.py
+++ b/src/realtime/src/realtime/_async/channel.py
@@ -501,13 +501,20 @@ class AsyncRealtimeChannel:
 
     def _broadcast_endpoint_url(self):
         return f"{http_endpoint_url(self.socket.http_endpoint)}/api/broadcast"
-
+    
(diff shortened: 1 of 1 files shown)


o-santi:
Closing as solved by #1346. The solution is actually similar, but the explanation for why it happens is not quite right. It isn't a race condition, but a quirk on how the message acks are processed.

PR #1339, opened 2025-12-18
fix(realtime): use 'event' instead of 'events' in postgres_changes protocol

### LINKED ISSUE #1319: Watching for postgres changes always times out
### Describe the bug

When subscribing to database changes using supabase-py, I see no events, until after about 10 seconds when I get `Channel subscription state: RealtimeSubscribeStates.TIMED_OUT`.

To check that it's a python problem and not something else, I tried this js code which works fine:

```
import { createClient } from '@supabase/supabase-js'

const supabase = createClient(
  process.env.SUPABASE_URL,
  process.env.SUPABASE_SERVICE_ROLE_KEY
)

const channel = supabase
  .channel('schema-db-changes')
  .on(
    'postgres_changes',
    {
      event: '*',
      schema: 'public',
    },
    (payload) => console.log(payload)
  )
  .subscribe()
```

### Reproduction

```python
import asyncio
import supabase
import os


async def watch_db(url, key):
    client = await supabase.acreate_client(url, key)

    def handle_payload(payload):
        print(payload)

    def subscribe_callback(state, error):
        if error:
            print(f"Error subscribing to channel: {error}")
        else:
            print(f"Channel subscription state: {state}")

    channel = client.channel("schema-db-changes")
    channel.on_postgres_changes("*", schema="public", callback=handle_payload)
    await channel.subscribe(subscribe_callback)

    await asyncio.Event().wait()


if __name__ == "__main__":
    asyncio.run(
        watch_db(os.environ["SUPABASE_URL"], os.environ["SUPABASE_SERVICE_ROLE_KEY"])
    )
```

### Steps to reproduce

1. Set env vars `SUPABASE_URL` and `SUPABASE_SERVICE_ROLE_KEY` to point to a locally running Supabase instance, mine has supabase/realtime:v2.65.2
2. Install supabase-py 2.24.0 (I also tried with 2.23.2 and got the same result)
3. Run the python code
4. Do some DB insertions

Expected: I first see a SUBCRIBED event and then I see the db events

Observed: I see no events, until after about 10 seconds when I get `Channel subscription state: RealtimeSubscribeStates.TIMED_OUT`

### Library affected

supabase-py

### Library version

supabase 2.24.0; supabase/realtime:v2.65.2

### Python version

Python 3.13.9; Python 3.14.0 on MacOS Tahoe 26.1

alexwhitmore:

Bug fix

Fixes #1319

`postgres_changes` subscriptions always time out after 10 seconds and never receive payloads. The client sends `events` (plural) but the server expects `event` (singular), causing a protocol mismatch in three places:

1. `binding_filter` sends `{"events": "*", "table": null}` instead of `{"event": "*"}`
2. `PostgresRowChange` model expects `events` field but server returns `event`
3. `channel.py` compares `server_binding.events` which doesn't exist

## What is the new behavior?

Subscriptions work correctly - state becomes `SUBSCRIBED` and payloads are received:
```
[2025-12-18 15:34:44.153631] Channel subscription state: RealtimeSubscribeStates.SUBSCRIBED

[2025-12-18 15:35:05.561223] *** RECEIVED PAYLOAD ***
{'data': {'schema': 'public', 'table': 'test_messages', 'commit_timestamp': '2025-12-18T14:35:05.504Z', 'type': <RealtimePostgresChangesListenEvent.Insert: 'INSERT'>, 'errors': None, 'columns': [{'name': 'id', 'type': 'int4'}, {'name': 'content', 'type': 'text'}, {'name': 'created_at', 'type': 'timestamptz'}], 'record': {'id': 16, 'content': 'testing fix', 'created_at': '2025-12-18T14:35:05.502162+00:00'}}, 'ids': [74307548]}
```

## Additional context

Tested with local Supabase (realtime v2.67.2), Python 3.13.3, supabase-py 1.2.0.

Changes:
- `types.py`: `binding_filter` sends `"event"` key, only includes `"table"` if specified
- `message.py`: `PostgresRowChange.events` → `event`, `table` is now optional  
- `channel.py`: `server_binding.events` → `server_binding.event`

### DIFF
--- a/src/realtime/src/realtime/types.py
+++ b/src/realtime/src/realtime/types.py
@@ -147,7 +147,7 @@ class PostgresChangesCallback:
 
     @property
     def binding_filter(self) -> dict[str, Optional[str]]:
-        binding = {"event": self.event}
+        binding: dict[str, Optional[str]] = {"event": self.event}
         if self.schema:
             binding["schema"] = self.schema
         if self.table:


o-santi:
Thank you, this indeed seems to be a bug.

> supabase-py 1.2.0.

But this doesn't seem right, how did you test these changes? Also, there are some `mypy` complaints, but I'll fix that shortly. 

Please, make sure to run `make ci` in the monorepo root before opening a PR.

alexwhitmore:
> Please, make sure to run `make ci` in the monorepo root before opening a PR.

Thanks @o-santi for the fix, I'll make sure to run `make ci` before future PRs. Just pulled in these changes locally and confirmed subscriptions still work correctly against local Supabase while satisfying mypy.

PR #1386, opened 2026-02-08
fix(realtime): wait for subscribe ack before returning

### LINKED ISSUE #1209: AsyncRealtimeChannel.subscribe returns without waiting for an answer
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When `subscribe` is called, the channel object will send the payload through the websocket connection and instantly return, without waiting for an answer on whether `subscribe` was successful or not. This creates some unexpected behavior, as `callback` is not guaranteed to be called after `await channel.subscribe(callback)` returns which create non trivial race conditions. 

Because of this, even if the callback crashes the program in the case of failure to subscribe to a channel, API calls after `subscribe` returns are not guarantee to be in a valid subscribed state, as it may be that `callback` hasn't been called yet.

## To Reproduce

Create a `supabase start` local instance, then run the example given in the README.md:
```py
# test.py
import asyncio
from typing import Optional

from realtime import AsyncRealtimeClient, RealtimeSubscribeStates

async def main():
    REALTIME_URL = "ws://localhost:54321/realtime/v1"
    API_KEY = "{API_KEY}" # api key here

    socket = AsyncRealtimeClient(REALTIME_URL, API_KEY)
    channel = socket.channel("test-channel")
    
    def _on_subscribe(status: RealtimeSubscribeStates, err: Optional[Exception]):
        raise SystemExit(100)

    await channel.subscribe(_on_subscribe)

if __name__ == '__main__':
    asyncio.run(main())
    print('`main()` ran successfully')
```

Which results in:
```
$ python test.py
`main()` ran successfully
```
Even though the `callback` should've crashed the program instantly.

## Expected behavior

After `channel.subscribe(callback)` is `await`ed, it should ensure that `callback` is called by waiting for the answer inside the `subscribe` async function. Currently, the only way to ensure that it is called is to manually block after the `await` point with some `asyncio.sleep` calls, which is not only ugly but also requires you to emulate `timeout`s externally, even though `subscribe` does seem to have an internal timeout mechanic.

## System information

- OS: NixOS
- Version of supabase-py: 2.5.2

Dodothereal:
## Summary
- make `AsyncRealtimeChannel.subscribe()` wait for a definitive subscribe outcome before returning
- capture subscribe outcomes (`SUBSCRIBED`, `CHANNEL_ERROR`, `TIMED_OUT`) in an internal future
- propagate callback exceptions through `await subscribe(...)` instead of returning early
- add unit tests verifying: (1) subscribe does not return before ack, and (2) callback exceptions are raised to caller

## Why
Issue #1209 reports `await channel.subscribe(callback)` returning before the server reply is handled, causing race conditions. This change ensures await semantics match subscription acknowledgement lifecycle.

## Testing
- `uv run --package realtime pytest tests/test_channel.py tests/test_timer.py` (run in `src/realtime`)
- `make realtime.mypy`
- `uv run ruff check src/realtime/src/realtime/_async/channel.py src/realtime/tests/test_channel.py`

Closes #1209

### DIFF
--- a/src/realtime/src/realtime/_async/channel.py
+++ b/src/realtime/src/realtime/_async/channel.py
@@ -189,6 +189,24 @@ class AsyncRealtimeChannel:
                 "Tried to subscribe multiple times. 'subscribe' can only be called a single time per channel instance"
             )
         else:
+            subscribe_result: asyncio.Future[None] = (
+                asyncio.get_running_loop().create_future()
+            )
+
+            def complete_subscription(
+                state: RealtimeSubscribeStates, error: Optional[Exception]
+            ) -> None:
+                if callback:
+                    try:
+                        callback(state, error)
+                    except Exception as exc:
+                        if not subscribe_result.done():
+                            subscribe_result.set_exception(exc)
+                        return
+
+                if not subscribe_result.done():
+                    subscribe_result.set_result(None)
+
             config: RealtimeChannelConfig = self.params["config"]
             broadcast = config.get("broadcast")
             presence = config.get("presence") or RealtimeChannelPresenceConfig(
@@ -245,7 +263,7 @@ class AsyncRealtimeChannel:
                             new_postgres_bindings.append(postgres_callback)
                         else:
                             asyncio.create_task(self.unsubscribe())
-                            callback and callback(
+                            complete_subscription(
                                 RealtimeSubscribeStates.CHANNEL_ERROR,
                                 Exception(
                                     "mismatch between server and client bindings for postgres changes"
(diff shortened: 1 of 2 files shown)


Dodothereal:
@olirice @silentworks could you please review this bugfix for #1209 when you have time? Thanks.

o-santi:
Hey @Dodothereal, thanks for the PR. I'm afraid we cannot merge this as of now because it would be breaking for most clients. People write their clients assuming that it doesnt block nowadays, so their code would be most likely wrong after this.

I intend to fix this problem, together with some other ones in realtime, in the next v3 version (https://github.com/supabase/supabase-py/pull/1370). Feel free to ask any other improvement that may be a breaking change there too.

Issue #1251, opened 2025-10-08
v3.0 tracking issue


o-santi:
## v3.0 Breaking Changes

This issue is intended to track and discuss all breaking changes to come in `v3.0` before landing them in master, in order to not only evolve ideas but also see how impactful they could be to existing code bases.

### 1\. Single way to configure parameters for requests.

Since the introduction of httpx clients as parameters in [#1117](<https://github.com/supabase/supabase-py/issues/1117>), it is possible to configure the same parameter in different ways, which not only introduces complexity in merging parameters internally, but also have complicated non-trivial semantics. For instance, httpx clients only support configuration through mutation, which makes sharing them cause troubles (see [SDK-441](https://linear.app/supabase/issue/SDK-441/shared-httpxclient-causes-base-url-clobber-between-postgrest-and), [#1249](<https://github.com/supabase/supabase-py/issues/1249>)).

Instead, I believe that supabase should offer its own way of configuring parameters to requests in a single unified class, shared by all packages of the monorepo, even if there's duplication with httpx parameters, as its the only sane way of not duplicating while still retaining the correct immutable semantics,  for sharing configuration between different clients.

### 2\. Stop using `dict`/`TypedDict` for everything

Of course, `dict` is useful in some places, but they should absolutely not be used as parameters to functions, like configuration options, as they do are not good for handling/manipulation. `TypedDict`s are mostly syntax sugar, do not offer validation, and do not handle missing keys at all (`total=False` just implies all keys are `Optional`, ie. accept `None`). Furthermore, they're generally mutable, which is not great for typing and sharing.

Instead, dedicated structures should be in place to set up options. `pydantic.BaseModel` are strictly better alternatives for they handle missing keys correctly, handle validation, field accessing instead of indexing (`a.b` vs `a["b"]`), and do not have invariance issues (see [this stackoverflow answer for more info](<https://stackoverflow.com/questions/75967867/how-can-i-type-annotate-a-general-nested-typeddict>)).

### 3\. Async vs Sync

[...]

PR #1389, opened 2026-02-11
feat!: drop python 3.9 support

timonrieger:

**Maintenance / compatibility:** Drop Python 3.9 support and require Python 3.10+.

- The project supports Python 3.9+.
- Type hints use `typing` collection aliases (e.g. `Dict`, `List`) for compatibility with 3.9.

## What is the new behavior?

- **Minimum Python version is 3.10.** All `requires-python` and tooling (ruff, mypy) are updated to target 3.10.
- **Type hints use built-in generics** where applicable: `Dict`/`List` (and similar) are replaced with `dict`/`list` (e.g. `dict[str, str]`, `list[User]`) in line with [PEP 585 – Type Hinting Generics In Standard Collections](https://peps.python.org/pep-0585/). This removes the need for the parallel `typing` collection hierarchy and keeps annotations consistent with Python 3.9+ runtime behavior (and 3.10+ tooling).
- READMEs and docs that mentioned "Python >= 3.9" are updated to "Python >= 3.10".

No intended change in runtime behavio this is a version bump and typing/style cleanup.

### DIFF
--- a/src/auth/src/supabase_auth/_async/gotrue_client.py
+++ b/src/auth/src/supabase_auth/_async/gotrue_client.py
@@ -123,13 +123,6 @@ class AsyncGoTrueClient(AsyncGoTrueBaseAPI):
         if headers:
             extra_headers.update(headers)
 
-        if sys.version_info < (3, 10):
-            warn(
-                "Python versions below 3.10 are deprecated and will not be supported in future versions. Please upgrade to Python 3.10 or newer.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-
         AsyncGoTrueBaseAPI.__init__(
             self,
             url=url or GOTRUE_URL,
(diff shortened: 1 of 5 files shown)


timonrieger:
cc @o-santi #1251

o-santi:
Thank you tim, but as explained here https://github.com/supabase/supabase-py/issues/1251#issuecomment-3880611960, we only intend to drop the 3.9 requirement when merging v3. 

The PRs already merged into `v3` have already dropped 3.9 compatibility for `storage` and `functions`, and I intend to do the same to the other versions. Given that this will just cause merge conflicts, I'll close this PR.

timonrieger:
The PR is against the v3 branch, not against main. I understand that you try to reduce merge conflicts. Are you doing these changes yourself close before merging v3 into main? If so, from a contributor perspective it would be helpful to know that this branch is internal to the team.

PR #1390, opened 2026-02-11
refactor: replace typing_extensions imports with standard typing

timonrieger:
stacked on #1389 

replaces typing_extensions imports with standard typing where possible

kept this against v3 to reduce rebase pain. non-breaking but also no feature that needs to land in v2


o-santi:
See https://github.com/supabase/supabase-py/pull/1389#issuecomment-3884178574

PR #1391, opened 2026-02-11
refactor: lint codebase with ruff

timonrieger:
stacked on #1390

unmodified result of running `uvx ruff check --fix` from the project root


timonrieger:
kept this against v3 to reduce rebase pain. non-breaking but also no feature that needs to land in v2

o-santi:
See https://github.com/supabase/supabase-py/pull/1389#issuecomment-3884178574

Issue #1207, opened 2023-12-20
maybe_single throws an error when 0 rows are returned


levity:
# Bug report

## Describe the bug

[maybe_single](https://postgrest-py.readthedocs.io/en/latest/api/request_builders.html#postgrest.SyncSelectRequestBuilder.maybe_single) is supposed to return None when the query returns no rows. But now it throws an APIError:

```postgrest.exceptions.APIError: {'message': 'Missing response', 'code': '204', 'hint': 'Please check traceback of the code', 'details': "Postgrest couldn't retrieve response, please check traceback of the code. Please create an issue in `supabase-community/postgrest-py` if needed."}```

## To Reproduce

Run any query that returns 0 rows, e.g:

`postgrestClientInstance.from_('my_table').select('*').eq('content', 'nonexistent').maybe_single().execute()`

## Expected behavior

should return `APIResponse(data=None)` as in previous versions

## System information

- supabase 2.3.0
- postgrest 0.13.0
- PostgreSQL 15.1 (Ubuntu 15.1-1.pgdg20.04+1) on aarch64-unknown-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, 64-bit

DhenPadilla:
Wondering if there's an update on this?

Krishna-js45:
Thanks for the clarification. Understood that this is a breaking-change candidate and will be addressed as part of v3.0. Appreciate the context.

o-santi:
> Wondering if there's an update on this?

This will be fixed in the next major version. See https://github.com/supabase/supabase-py/pull/1370 for current status.

gostak-dd:
tldr maybe_single() asks if there's a match, gets a weird response it doesn't understand, and silently breaks down. much like me on Hinge...

o-santi:
I believe I found a very simple solution that is not a breaking change, in #1424. I'll merge it for now, while we don't have the next major version ready for release.

PR #1393, opened 2026-02-11
fix(postgrest): handle empty result & PGRST116 in maybe_single correctly

kelvinvelasquez-SDE:
# Fix: maybe_single() unexpected error on 0 rows (Issue #1207)

## Description
This PR fixes an issue where `maybe_single()` would raise an exception (often `APIError: Missing response` or similar) when PostgREST returns 0 rows, instead of returning `None` as documented and expected.

The issue stems from:
1.  Handling of `PGRST116` error: The previous check for "The result contains 0 rows" string was brittle and missed variations (e.g. from newer PostgREST versions or different configurations).
2.  Response handling: A successful 200 OK response with an empty list `[]` was returned wrapped in `SingleAPIResponse(data=[])` instead of `None`, which is semantically incorrect for `maybe_single`.

## Changes
- Updated `AsyncMaybeSingleRequestBuilder.execute` and `SyncMaybeSingleRequestBuilder.execute` in `src/postgrest`.
- Added robust check for `PGRST116` error code combined with "0 rows" or "no rows" in details/message.
- Added check for empty list data (`[]`) in successful responses to return `None`.
- Ensured that unhandled exceptions are re-raised correctly instead of being swallowed or converted to generic errors.

## Verification
Reproduction script attached confirms:
- **Case 1 (200 OK, `[]`)**: Now returns `None` (was `data=[]`).
- **Case 2 (406 Error, Standard)**: Returns `None` (Success).
- **Case 3 (406 Error, Variant)**: Returns `None` (was raising Exception).
- **Case 4 (Other Error)**: Correctly raises `APIError`.

## Checklist
- [x] Code follows the style guidelines of this project
- [x] Performed a self-review of my own code
- [x] Commented my code, particularly in hard-to-understand areas
- [x] Changes generate no new warnings
- [x] Added reproduction script/test case

## Summary by CodeRabbit

* **Bug Fixes**
  * MaybeSingle now consistently returns None for empty query results and for API responses that explicitly indicate "0 rows" or "no rows" instead of raising an error.
  * Improved handling for HEAD responses when using maybe_single().

* **Tests**
  * Added comprehensive synchronous and asynchronous tests covering empty results, specific API error cases, and HEAD behavior.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,3 +43,10 @@ ignore = ["F403", "E501", "E402", "UP006", "UP035"]
 [tool.ruff.lint.pyupgrade]
 # Preserve types, even if a file imports `from __future__ import annotations`.
 keep-runtime-typing = true
+
+[dependency-groups]
+dev = [
+    "aioresponses>=0.7.8",
+    "responses>=0.25.8",
+    "respx>=0.22.0",
+]
--- a/src/postgrest/run-unasync.py
+++ b/src/postgrest/run-unasync.py
@@ -5,7 +5,15 @@ import unasync
 paths = Path("src/postgrest").glob("**/*.py")
 tests = Path("tests").glob("**/*.py")
 
-rules = (unasync._DEFAULT_RULE,)
+rules = [
+    unasync.Rule(
+        fromdir="/_async/",
+        todir="/_sync/",
+        additional_replacements={
+            "AsyncClient": "Client",
+        },
+    )
+]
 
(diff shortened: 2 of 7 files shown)


o-santi:
Sadly this cannot land on the current version because it would be a breaking change. See https://github.com/supabase/supabase-py/issues/1207#issuecomment-3210453495 for more information.

I intend to fix this issue in the next major version, together with other stuff.

PR #1444, opened 2026-04-06
fix(json): add datetime and UUID to JSON simple types

### LINKED ISSUE #1443: `insert` does not expect TypedDict provided by generated types
### Describe the bug

Following the guidance here:

```
# Insert
                                          .execute().data]
```

Image

Image

Image

### Reproduction

```python
Included above
```

### Steps to reproduce

_No response_

### Library affected

supabase-py

### Library version

2.28.3

### Python version

3.13.0

o-santi:

Fixes #1443, adding `datetime`, `date`, `time` and `uuid` as valid `JSON` types regarding `postgrest`, in order to support generated classes that have these types as fields.

### DIFF
--- a/src/postgrest/src/postgrest/types.py
+++ b/src/postgrest/src/postgrest/types.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Mapping, Sequence
+from datetime import date, datetime, time
 from typing import Union
+from uuid import UUID
 
 from httpx import AsyncClient, BasicAuth, Client, Headers, QueryParams
 from pydantic import TypeAdapter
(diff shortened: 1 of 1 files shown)


microfauna:
Ah worth noting, I am still getting 

```
TypeError: Object of type datetime is not JSON serializable
``` 
with this change

Since this seems to root all the way down to `httpx` library's code (it needs `default=str` as an arg) we can maybe dump the json way earlier, in postgrest's code (maybe using JSONAdapter)

image

mandarini (approved):

o-santi:
@microfauna if `json.dumps` doesn't serialize `datetime`, then this PR doesn't make sense. In the `v3` branch, we are already following the new approach of doing the JSON serialization ourselves, so the best course of action is to just wait for `v3` to come out, because I'd need to backport most of the changes from that branch into main, and that would be a considerable amount of work.

microfauna:
Ah I was running it with the changes made to my local version - so the PR may still make sense. I'd just say to to try to test it with a datetime

PR #1450, opened 2026-04-10
fix: updated type annotations for insert methods to support TypedDict

### LINKED ISSUE #1443: `insert` does not expect TypedDict provided by generated types

SUSH9391:
## Description

Resolves #1443

The `insert()`, `upsert()`, and `update()` methods only accepted plain
`dict` in their type annotations. Supabase's official CLI generates
`TypedDict` classes for database tables and the docs encourage users to
pass them directly to these methods. However type checkers (mypy/pyright)
rejected this with a type error even though it works fine at runtime.

**Example of the problem:**
```python
class MovieInsert(TypedDict):
    name: str

# Type checker incorrectly rejects this even though it works at runtime
movies.insert(MovieInsert(name="foo")).execute()
```

## Fix

Changed the type annotations for payload arguments to use
`Union[Mapping[str, Any], Sequence[Mapping[str, Any]]]`, which correctly
covers plain dicts, lists of dicts, and TypedDict instances.

Files updated:
- `src/postgrest/base_request_builder.py` (`pre_insert`, `pre_update`, `pre_upsert`)
- `src/postgrest/_sync/request_builder.py` (`insert`, `update`, `upsert`)
- `src/postgrest/_async/request_builder.py` (`insert`, `update`, `upsert`)

## Tests

- Added `tests/test_typed_dict_insert.py` verifying TypedDict instances
  are accepted by `insert()` for both single rows and lists
- `make postgrest.mypy` passes cleanly
- `uv run pytest tests/_sync/test_request_builder.py tests/_async/test_request_builder.py -v`

### DIFF
--- a/src/postgrest/src/postgrest/_async/request_builder.py
+++ b/src/postgrest/src/postgrest/_async/request_builder.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Generic, Literal, Optional, TypeVar, Union, overload
+from typing import Any, Generic, Literal, Optional, TypeVar, Union, overload, Mapping, Sequence
 
 from httpx import AsyncClient, BasicAuth, Headers, QueryParams, Response
 from pydantic import ValidationError
@@ -271,7 +271,7 @@ class AsyncRequestBuilder:  #
 
     def insert(
         self,
-        json: JSON,
+        json: Union[Mapping[str, Any], Sequence[Mapping[str, Any]]],
         *,
         count: Optional[CountMethod] = None,
         returning: ReturnMethod = ReturnMethod.representation,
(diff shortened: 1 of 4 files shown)


o-santi:
This does not fix the issue, because the underlying problem is that `json.dumps` does not support `uuid` or `datetime`.

SUSH9391:
Thanks for the clarification. So the fix should be in the JSON serialization layer to handle uuid and datetime types from generated TypedDicts — not just the type annotations. I'll look into this.

Issue #1455, opened 2026-04-13
ValidationError when creating signed URLs for non-existent files


NafeesMadni:
### Describe the bug

* **What I am doing**: I am using the `create_signed_urls` function to generate signed URLs for files in a Supabase storage bucket.
* **What I expect is**: If a specified file path does not exist in the bucket, I expect the function to gracefully handle the API response (which may return `None` for the `signedURL` for that file) without raising a `ValidationError`.
* **What actually happening is**: When a file path does not exist, the `create_signed_urls` function raises a `pydantic_core._pydantic_core.ValidationError` because the Pydantic model `SignedUrlsJsonItem` expects a string type for `signedURL`, but the API returns None (null) in such cases.

### Reproduction
  When calling `create_signed_urls` with a list of file paths, if one of the files does not exist, the Supabase API returns a response similar to this:
```json
[
    {
        "error": null,
       "path": "existing_file.txt",
       "signedURL": "/object/sign/fileio/existing_file.txt?token=..."
    },
    {
       "error": "Either the object does not exist or you do not have access to it",
       "path": "non_existent_file.txt",
       "signedURL": null
   }
]
```
  The `SignedUrlsJsonResponse` in the current SDK is defined using `SignedUrlsJsonItem`, which currently expects signedURL to be a string:
```py
class SignedUrlsJsonItem(BaseModel, extra="ignore"):
   error: Optional[str]
   path: str
   signedURL: str # This is the problematic field, should be Optional[str]
   
SignedUrlsJsonResponse = TypeAdapter(list[SignedUrlsJsonItem])
```
Because the SDK definition requires `signedURL` to be a string, it fails validation when it receives null from the API, leading to the 
```bash
pydantic_core._pydantic_core.ValidationError: 1 validation error for list[SignedUrlsJsonItem] 1.signedURL Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

  Here's an example of how to trigger this (assuming non_existent_file.txt does not exist):

```python
# Assuming you have a client and bucket initialized
# from supabase import create_client, Client

# url: str = "YOUR_SUPABASE_URL"
# key: str = "YOUR_SUPABASE_ANON_KEY"
# supabase: Client = create_client(url, key)
# bucket_name = "your-bucket-name"

# Example call that might trigger the error if 'non_existent_file.txt' does not exist
# The actual error occurs within the SDK's internal processing of the API response.
try:
   # Example: a list of file paths, where one or more might not exist
[...]

PR #1456, opened 2026-04-13
fix(storage3): Handle None for non-existent signed URLs

NafeesMadni:

Bug fix

The `create_signed_urls` function raises a `pydantic_core._pydantic_core.ValidationError` when attempting to generate signed URLs for files that do not exist in a Supabase storage bucket. This occurs because the Supabase API returns null for the `signedURL` field for non-existent files, but the SDK's `SignedUrlsJsonItem` Pydantic model expects a string.

**Relevant issue**: [#1455](https://github.com/supabase/supabase-py/issues/1455)

## What is the new behavior?

This PR updates the `SignedUrlsJsonItem` Pydantic model to correctly mark the `signedURL` field as `Optional[str]`. Additionally, the `create_signed_urls` methods in both the asynchronous and synchronous file API implementations have been modified to gracefully handle `None` values for `signedURL` from the API response.

Now, if a requested file does not exist, the corresponding entry in the `CreateSignedUrlResponse` list will correctly reflect `None` for `signedURL` and `signedUrl`, without causing a validation error.

## Additional context

This fix enhances the robustness of the storage client, preventing crashes when querying for signed URLs of potentially non-existent files. Users can now safely request signed URLs for a list of files, even if some are missing, without the application crashing due to Pydantic validation failures.

### DIFF
--- a/src/storage/src/storage3/_async/file_api.py
+++ b/src/storage/src/storage3/_async/file_api.py
@@ -273,14 +273,23 @@ class AsyncBucketActionsMixin:
         data = SignedUrlsJsonResponse.validate_json(response.content)
         signed_urls = []
         for item in data:
-            # Prepare URL
-            url = self._make_signed_url(item.signedURL, download_query)
-            signed_item: CreateSignedUrlResponse = {
-                "error": item.error,
-                "path": item.path,
-                "signedURL": url["signedURL"],
-                "signedUrl": url["signedURL"],
-            }
+            signed_url: Optional[str] = item.signedURL
+            if signed_url:
+                # Prepare URL
+                url = self._make_signed_url(signed_url, download_query)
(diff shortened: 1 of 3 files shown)


o-santi:
This PR also needs to fix the tests and some annotations in order to keep `make mypy` from not throwing an error. 

I'll close it in favor of https://github.com/supabase/supabase-py/pull/1461, where I've applied a very similar fix to it, while keeping the types correct.

NafeesMadni:
@o-santi Thank you for the feedback! I'm glad the logic was on the right track. I really appreciate you pointing out the mypy requirements. I’ll take a look at [#1461](https://github.com/supabase/supabase-py/pull/1461) to see how you handled the annotations so I can keep that in mind for future PRs. Happy to see the fix getting merged!

PR #1465, opened 2026-04-17
Fix ValidationError for null signedURL in create_signed_urls

### LINKED ISSUE #1455: ValidationError when creating signed URLs for non-existent files

fuleinist:
Fixes ValidationError when create_signed_urls is called with non-existent file paths. Changed signedURL: str to signedURL: Optional[str] in SignedUrlsJsonItem class. Fixes #1455

### DIFF
--- a/src/storage/src/storage3/types.py
+++ b/src/storage/src/storage3/types.py
@@ -170,7 +170,7 @@ class SignedUrlJsonResponse(BaseModel, extra="ignore"):
 class SignedUrlsJsonItem(BaseModel, extra="ignore"):
     error: Optional[str]
     path: str
-    signedURL: str
+    signedURL: Optional[str]
 
 
 SignedUrlsJsonResponse = TypeAdapter(list[SignedUrlsJsonItem])


o-santi:
Not needed. Already solved by https://github.com/supabase/supabase-py/pull/1461.

PR #1480, opened 2026-04-29
feat(auth): add third party auth support through access_token

o-santi:

Add support for third party authentication, by passing in an `access_token` function that provides an external id token. 

Based on https://github.com/supabase/supabase-js/pull/1004.


grdsdev (changes requested):
This implementation fetches the access token only on client creation? The feature should be to always requests token when a sub-package makes a request. In the JS it fetches the token inside the fetch implementation.

grdsdev (approved):

PR #1483, opened 2026-05-05
fix(postgrest): serialize generated typed dict payloads (#1443)

### LINKED ISSUE #1443: `insert` does not expect TypedDict provided by generated types

SM1LE-X:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Issue
Fixes #1443

Generated Python REST types can produce `TypedDict` write payloads with PostgreSQL-derived Python values such as `UUID`, `datetime`, `date`, and `time`.

The current `insert`, `upsert`, and `update` signatures only accept the recursive response `JSON` type, so generated `TypedDict` payloads are rejected by type checkers. Widening the annotation alone does not fix the runtime path because request execution still relies on stdlib JSON serialization through HTTPX, which raises `TypeError` for `UUID` and `datetime` values.

## What is the new behavior?
This PR keeps response `JSON` parsing strict and adds a separate write payload type for row-shaped request bodies.

Write methods now accept generated-style payloads, and PostgREST serializes supported generated values before sending the HTTP request body:

* `UUID` is serialized with `str(value)`.
* `datetime`, `date`, and `time` are serialized with `value.isoformat()`.
* Existing HTTPX JSON behavior is preserved: compact separators, UTF-8 output, `Content-Type: application/json` when absent, and `allow_nan=False`.

**Changes made:**
* Added `WriteJSON` for `Mapping[str, Any] | Sequence[Mapping[str, Any]]` write payloads.
* Updated sync and async `insert`, `upsert`, and `update` to use `WriteJSON`.
* Encoded request JSON in PostgREST before calling HTTPX so generated `TypedDict` values serialize correctly at runtime.
* Added sync and async tests for generated-style `TypedDict` payloads containing `UUID`, `datetime`, `date`, and `time`.
* Added coverage for preserving an existing `Content-Type` and rejecting non-finite floats.

## Verification
* `uv run --package postgrest mypy src/postgrest tests`
* `uv run --package postgrest pytest --cov=./ --cov-report=xml -vv` - 292 passed
* `uv run ruff check <touched files>`
* `uv run ruff format --check <touched files>`
* Pyright probe for generated-style `TypedDict` payloads - 0 errors

## Checklist:
- [x] I have read the [CONTRIBUTING](https://github.com/supabase/supabase-py/blob/main/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.

### DIFF
--- a/src/postgrest/src/postgrest/_async/request_builder.py
+++ b/src/postgrest/src/postgrest/_async/request_builder.py
@@ -23,7 +23,7 @@ from ..base_request_builder import (
     pre_upsert,
 )
 from ..exceptions import APIError, APIErrorFromJSON, generate_default_error_message
-from ..types import JSON, ReturnMethod
+from ..types import ReturnMethod, WriteJSON
 from ..utils import model_validate_json
 
 ReqConfig = RequestConfig[AsyncClient]
@@ -326,7 +326,7 @@ class AsyncRequestBuilder:  #
 
     def insert(
         self,
-        json: JSON,
+        json: WriteJSON,
         *,
         count: Optional[CountMethod] = None,
         returning: ReturnMethod = ReturnMethod.representation,
@@ -367,7 +367,7 @@ class AsyncRequestBuilder:  #
 
     def upsert(
         self,
-        json: JSON,
+        json: WriteJSON,
         *,
         count: Optional[CountMethod] = None,
         returning: ReturnMethod = ReturnMethod.representation,
@@ -412,7 +412,7 @@ class AsyncRequestBuilder:  #
 
     def update(
         self,
-        json: JSON,
+        json: WriteJSON,
         *,
         count: Optional[CountMethod] = None,
         returning: ReturnMethod = ReturnMethod.representation,
--- a/src/postgrest/src/postgrest/_sync/request_builder.py
+++ b/src/postgrest/src/postgrest/_sync/request_builder.py
@@ -23,7 +23,7 @@ from ..base_request_builder import (
     pre_upsert,
 )
 from ..exceptions import APIError, APIErrorFromJSON, generate_default_error_message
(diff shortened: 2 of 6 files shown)


SM1LE-X:
Addressed the write-surface coverage gap and tightened the payload alias:

- added sync and async regression tests for generated `TypedDict` serialization through `upsert`
- added sync and async regression tests for generated `TypedDict` serialization through `update`
- kept the shared serializer guard covered with an unsupported-object test so arbitrary objects still fail before transport
- replaced the `Any` write-payload values with `object`; `TypedDict` is exposed to type checkers as `Mapping[str, object]`, so this keeps generated row types assignable without using `Any`

Verification after the update:

- focused serialization tests: 7 passed
- full PostgREST tests: 292 passed
- mypy: no issues in 30 source files
- ruff check/format on touched files: clean
- Pyright generated-`TypedDict` probe for sync/async `insert`, `upsert`, and `update`: 0 errors

o-santi:
To be honest, this is an interesting solution, but ultimately I think it is the wrong way to solve it. We shouldn't rely on `json.loads` (or `httpx`'s `.json()`, which uses `loads` under the hood) because it returns `Any`, so it is horrible for type checking purposes.

In the `v3` branch, this issue is already solved by using a pydantic model to parse the `JSON` type, giving us not only more performance, but also full type safety. In that case, adding the correct types to the recursive `JSON` type already fixes the issue. I would much appreciate contributions there instead.

PR #1487, opened 2026-05-12
Speed up postgrest JSON validation

### LINKED ISSUE #1486: `postgrest.types.JSONAdapter` is much slower than `TypeAdapter(JsonValue)`
### Describe the bug

Hi, I noticed that the JSON alias currently used by `postgrest-py`:

```python
JSON = TypeAliasType(
    "JSON",
    "Union[None, bool, str, int, float, Sequence[JSON], Mapping[str, JSON]]",
)
```

is much slower (13x) than Pydantic's built-in [`JsonValue`](https://pydantic.dev/docs/validation/2.11/api/pydantic/types/#pydantic.types.JsonValue) on the same payload.

### Reproduction

```python
import json
import statistics
import timeit

from postgrest.types import JSONAdapter
from pydantic import TypeAdapter
from pydantic.types import JsonValue

postgrest_json = JSONAdapter
json_value = TypeAdapter(JsonValue)

payload = json.dumps(
    [
        {
            "id": i,
            "name": f"doc-{i}",
            "labels": ["a", "b", "c"],
            "status": {"id": 1, "label": "Review", "done": False},
            "metadata": {"scores": [0.9, 0.8, 0.7], "flags": {"pinned": True, "archived": False}},
        }
        for i in range(1_000)
    ]
).encode()


def bench(label: str, stmt: str) -> float:
    ms = statistics.mean(timeit.repeat(stmt, number=1, repeat=5, globals=globals())) * 1000
    print(f"{label:24} {ms:8.2f} ms")
    return ms


print(f"payload_size={len(payload):,} bytes")
bench("postgrest TypeAliasType", "postgrest_json.validate_json(payload)")
bench("pydantic.JsonValue", "json_value.validate_json(payload)")
bench("json.loads", "json.loads(payload)")

On my machine this is the output:


payload_size=200,780 bytes
postgrest TypeAliasType     11.65 ms
pydantic.JsonValue           0.86 ms
json.loads                   0.76 ms
```

### Steps to reproduce

_No response_

### Library affected

postgrest

### Library version

postgrest 2.29.0

### Python version

3.14.0

snoopuppy582:
Fixes #1486

## Summary
- switch `postgrest.types.JSONAdapter` to Pydantic's optimized `JsonValue` schema
- keep the public `JSON` type alias unchanged for request typing
- add regression coverage that guards the `JsonValue` adapter path and nested JSON byte validation

## Local validation
- Before fix benchmark from the issue payload: `JSONAdapter.validate_json(payload)` averaged `44.78 ms`; `TypeAdapter(JsonValue)` averaged `3.93 ms`
- After fix benchmark from the issue payload: `JSONAdapter.validate_json(payload)` averaged `3.56 ms`; `TypeAdapter(JsonValue)` averaged `3.74 ms`
- `uv run --project src/postgrest pytest src/postgrest/tests/test_types.py -q`
- `uv run --project src/postgrest pytest src/postgrest/tests/test_types.py src/postgrest/tests/test_utils.py src/postgrest/tests/_sync/test_request_builder.py src/postgrest/tests/_async/test_request_builder.py -q`
- `uv run --project src/postgrest ruff check src/postgrest/src/postgrest/types.py src/postgrest/tests/test_types.py`

## Notes
- I tried the full `src/postgrest/tests` suite locally, but it did not finish before the 4-minute timeout in my Windows environment. The targeted request-builder/type tests above completed successfully.
- `uv run --locked` currently reports that the existing workspace lockfile needs an update, so the PR intentionally does not include lockfile changes.

### DIFF
--- a/src/postgrest/src/postgrest/types.py
+++ b/src/postgrest/src/postgrest/types.py
@@ -6,6 +6,7 @@ from typing import Union
 
 from httpx import AsyncClient, BasicAuth, Client, Headers, QueryParams
 from pydantic import TypeAdapter
+from pydantic.types import JsonValue
 from typing_extensions import TypeAliasType
 from yarl import URL
 
@@ -18,7 +19,7 @@ else:
 JSON = TypeAliasType(
     "JSON", "Union[None, bool, str, int, float, Sequence[JSON], Mapping[str, JSON]]"
 )
-JSONAdapter: TypeAdapter = TypeAdapter(JSON)
+JSONAdapter: TypeAdapter = TypeAdapter(JsonValue)
 
 
 class CountMethod(StrEnum):
--- a/src/postgrest/tests/test_types.py
+++ b/src/postgrest/tests/test_types.py
@@ -1,10 +1,11 @@
+from pydantic import TypeAdapter
+from pydantic.types import JsonValue
+
 from postgrest.types import JSONAdapter
 
 
-def test_json_adapter_uses_pydantic_json_value_schema():
-    assert JSONAdapter.core_schema["schema"]["schema_ref"].startswith(
-        "pydantic.types.JsonValue:"
-    )
+def test_json_adapter_schema_matches_pydantic_json_value():
+    assert JSONAdapter.core_schema == TypeAdapter(JsonValue).core_schema
 
 
 def test_json_adapter_validates_nested_json_bytes():
--- a/src/postgrest/tests/test_types.py
+++ b/src/postgrest/tests/test_types.py
@@ -1,0 +1,24 @@
+from postgrest.types import JSONAdapter
+
+
+def test_json_adapter_uses_pydantic_json_value_schema():
+    assert JSONAdapter.core_schema["schema"]["schema_ref"].startswith(
+        "pydantic.types.JsonValue:"
+    )
+
+
+def test_json_adapter_validates_nested_json_bytes():
+    payload = (
+        b'{"items":[{"id":1,"tags":["a","b"],'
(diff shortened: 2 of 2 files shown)


AmaLS367 (commented):
Thanks for this PR — the performance improvement is impressive and the change is clean. The core fix looks correct.

One suggestion on the test `test_json_adapter_uses_pydantic_json_value_schema`:

```python
assert JSONAdapter.core_schema["schema"]["schema_ref"].startswith(
    "pydantic.types.JsonValue:"
)
```

The intent is good (guarding the fast path), but the nested `["schema"]["schema_ref"]` path and the hardcoded `"pydantic.types.JsonValue:"` string tie the test to Pydantic's internal schema structure. A more resilient alternative would be to compare against a reference adapter:

```python
def test_json_adapter_schema_matches_json_value():
    from pydantic.types import JsonValue as PydanticJsonValue
    from pydantic import TypeAdapter
    assert JSONAdapter.core_schema == TypeAdapter(PydanticJsonValue).core_schema
```

This guards the same regression (someone reverting to the slow `TypeAliasType` path) without hardcoding internal schema details. Just a suggestion though — not a blocker.

snoopuppy582:
Thanks for the suggestion. I updated the test to compare `JSONAdapter.core_schema` against `TypeAdapter(JsonValue).core_schema` instead of asserting the nested Pydantic schema ref string.

Verified locally:
- `uv run pytest tests/test_types.py -q` -> 2 passed

o-santi:
This is already going to be changed in `v3`. Also, I don't think it really is faster since `JsonValue` is also just a recursive `TypeAdapter`.

PR #1514, opened 2026-06-10
Fix JSON type alias for datetime and UUID in TypedDicts

### LINKED ISSUE #1443: `insert` does not expect TypedDict provided by generated types

aydgnme:
## Problem
The `JSON` type alias in `postgrest/types.py` did not include `datetime`, 
`date`, `time`, or `UUID`. The Supabase CLI generates TypedDicts that 
use these types, causing type errors when passing them to `.insert()`, 
`.upsert()`, or `.update()`.

## Fix
Added `datetime`, `date`, `time`, and `UUID` to the `JSON` type alias. 
These types are already serializable by httpx/pydantic — this was purely 
a type annotation gap.

## Testing
All 43 existing sync unit tests pass.

Fixes #1443

### DIFF
--- a/src/postgrest/src/postgrest/types.py
+++ b/src/postgrest/src/postgrest/types.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Mapping, Sequence
+from datetime import date, datetime, time
 from typing import Union
+from uuid import UUID
 
 from httpx import AsyncClient, BasicAuth, Client, Headers, QueryParams
 from pydantic import TypeAdapter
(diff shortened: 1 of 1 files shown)


o-santi:
> These types are already serializable by httpx/pydantic — this was purely a type annotation gap.

This is wrong, and this PR does not fix the underlying issue. `httpx` uses `json.dumps` to serialize these, which does not know how to handle `date`, `time` or `UUID`. In order to fix this, we'd need to stop using httpx's json serialization and start serializing it on our own. This is already done in v3, so I'm just avoiding re-doing it again to not duplicate work.

PR #1520, opened 2026-06-26
docs: add SECURITY.md

spydon:
## What

Adds a `SECURITY.md` to the repository root, modeled on the one in the main [supabase/supabase](https://github.com/supabase/supabase/blob/master/SECURITY.md) repository.

## Why

The repository had no security policy. GitHub surfaces `SECURITY.md` in the Security tab and links to it when users want to report a vulnerability, so this gives security researchers clear guidance on how to responsibly disclose issues. Reports are directed to the HackerOne VDP, and the file covers scope, testing, reporting, and disclosure guidelines.

Closes SDK-1099

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,47 @@
+Contact: https://hackerone.com/supabase
+Canonical: https://supabase.com/.well-known/security.txt
+
+
+At Supabase, we consider the security of our systems a top priority. But no matter how much effort we put into system security, there can still be vulnerabilities present.
+
+If you discover a vulnerability, we would like to know about it so we can take steps to address it as quickly as possible. We would like to ask you to help us better protect our clients and our systems.
+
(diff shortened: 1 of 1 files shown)


Tr00d (approved):

o-santi:
I don't think we should be duplicating this file in every repository. I'm sure any security researcher can find the main repo's file just fine.

spydon:
> I don't think we should be duplicating this file in every repository. I'm sure any security researcher can find the main repo's file just fine.

It's good practice to expose it everywhere where it is needed, but you have a point, we can just set it in the `.github` repo instead so that we get a default for all repos from there instead.
https://github.com/supabase/.github/pull/21

spydon:
Quintin beat me to it 😄 
https://github.com/supabase/.github/pull/20

PR #1522, opened 2026-06-29
fix: httpx deprecation warnings

jpfeuffer:

Bug fix: remove deprecation warnings.

```
.../python3.12/site-packages/supabase/_sync/client.py:302: DeprecationWarning: The 'timeout' parameter is deprecated. Please configure it in the http client instead.
  return SyncPostgrestClient(
.../python3.12/site-packages/supabase/_sync/client.py:302: DeprecationWarning: The 'verify' parameter is deprecated. Please configure it in the http client instead.
  return SyncPostgrestClient(
```

etc.

## What is the new behavior?

No warnings.

### DIFF
--- a/src/supabase/Makefile
+++ b/src/supabase/Makefile
@@ -23,7 +23,6 @@ unasync:
 	uv run run-unasync.py
 build-sync: unasync
 	sed -i 's/asyncio.create_task(self.realtime.set_auth(access_token))//g' src/supabase/_sync/client.py
-	sed -i '/^import asyncio$$/d' src/supabase/_sync/client.py
 	sed -i 's/asynch/synch/g' src/supabase/_sync/auth_client.py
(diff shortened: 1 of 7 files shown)


jpfeuffer:
@o-santi Could you have a look, please? It is a very simple change that get rids of long-standing annoying warnings and makes this whole library futureproof.

o-santi:
All httpx specific parameters are going away in the next major version, as the http client won't be hard coded in anymore. Additionally, there is no need to test for warnings being logged or not, please review the generated code from the LLM before opening a PR.

jpfeuffer:
Then when is that release coming? I think it would still make sense to fix those warnings especially since the changes are so minor and people will still be using 2.x for a while.

I left the tests intentionally because it would at least show in the first CI runs, that the warnings are gone and we could still remove afterwards.
This was important since you are obviously not paying attention to deprecation warnings in your CI.

PR #1530, opened 2026-07-08
fix(postgrest): coerce non-string values in array filters

### LINKED ISSUE #1529: postgrest: array filters (cs/cd/contains/contained_by/ov) raise TypeError on non-string values
## Bug

The array/range filter methods in the PostgREST query builder crash with `TypeError` when given a non-string iterable (e.g. an integer array), even though their signatures accept `Iterable[Any]`. Integer array columns (`int[]`) are common in Postgres, so `.contains("ids", [1, 2, 3])` is a natural call.

`in_` already handles this (it maps values through `sanitize_param`), but `cs`, `cd`, `contains`, `contained_by`, `ov` (and `overlaps`, which delegates to `ov`) do a bare `",".join(values)` and fail.

## Reproduce

```python
from httpx import Client
from yarl import URL
from postgrest import SyncFilterRequestBuilder
from postgrest._sync.request_builder import RequestConfig

b = SyncFilterRequestBuilder(RequestConfig(Client(), URL('/t'), 'GET', {}, {}, None, {}))
b.contains('x', [1, 2, 3])   # TypeError: sequence item 0: expected str instance, int found
```

`in_('x', [1, 2, 3])` works; `cs`/`cd`/`contains`/`contained_by`/`ov`/`overlaps` all raise.

## Expected

The values should be coerced to `str` (as `in_` effectively does), producing e.g. `x=cs.{1,2,3}`.

I have a fix + tests ready and will open a PR.

gottostartsomewhere:
## What

Fixes #1529.

The array/range filter methods `cs`, `cd`, `contains`, `contained_by`, and `ov` (and `overlaps`, which delegates to `ov`) passed values directly to `",".join(...)`, which raises `TypeError` on non-string iterables like integer arrays — even though the signatures accept `Iterable[Any]`. `in_` already avoids this by mapping through `sanitize_param`.

```python
builder.contains("ids", [1, 2, 3])
# TypeError: sequence item 0: expected str instance, int found
```

Integer array columns (`int[]`) are common in Postgres, so this is easy to hit.

## Change

Coerce each element with `str()` before joining in the five affected methods, so `.contains("ids", [1, 2, 3])` builds `{1,2,3}`. This mirrors what `in_` already does and leaves existing string behaviour unchanged (verified against the existing tests, including ones with reserved characters like `overlaps("x", ["is:closed", "severity:high"])`).

The change is in the shared `base_request_builder.py`, so it fixes both sync and async builders.

## Tests

Added `test_{contains,cs,cd,contained_by,overlaps}_non_string_items` to both the async and sync filter-builder test suites. They fail on `main` (`TypeError`) and pass with the fix. Full filter test suite: 84 passed.

### DIFF
--- a/src/postgrest/src/postgrest/base_request_builder.py
+++ b/src/postgrest/src/postgrest/base_request_builder.py
@@ -455,11 +455,11 @@ class BaseFilterRequestBuilder(Generic[C]):
         return self.filter(column, Filters.IN, f"({values})")
 
     def cs(self: Self, column: str, values: Iterable[Any]) -> Self:
-        values = ",".join(values)
+        values = ",".join(str(v) for v in values)
         return self.filter(column, Filters.CS, f"{{{values}}}")
 
     def cd(self: Self, column: str, values: Iterable[Any]) -> Self:
-        values = ",".join(values)
+        values = ",".join(str(v) for v in values)
         return self.filter(column, Filters.CD, f"{{{values}}}")
 
     def contains(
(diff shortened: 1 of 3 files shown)


o-santi:
I'll accept this as a patch up but I think the proper solution would be to remove the `Any` from postgrest's argument everywhere, and be more strict on what types the user can/should provide.

o-santi (approved):

PR #1535, opened 2026-07-10
fix(postgrest): normalize boolean values in is_() to lowercase literals

winklemad:
No linked issue — found by reading the postgrest filter builder.

Bug fix.

`is_()` normalizes Python `None` to the PostgREST literal `null`, but does nothing for Python booleans. A boolean therefore falls through to `filter()`, which formats it into the query string as `is.True` / `is.False`:

```python
client.table("t").select("*").is_("active", True)
# -> ?active=is.True
```

PostgREST's `is` operator is case-sensitive and only accepts the lowercase trilean literals `null`, `true`, `false`, `unknown` (see postgrest/postgrest#2077). `is.True` is rejected by the server, so `.is_(col, True)` / `.is_(col, False)` — the natural way to filter a boolean column against `IS TRUE` / `IS FALSE` — produce a failing request.

## What is the new behavior?

`is_()` converts `True`/`False` to `"true"`/`"false"`, using `is True` / `is False` identity checks so only real booleans are touched — strings such as `"unknown"` / `"not_null"` and any other value pass through unchanged. This matches supabase-js, whose `.is('col', true)` emits `is.true`.

```python
client.table("t").select("*").is_("active", True)   # -> ?active=is.true
client.table("t").select("*").is_("active", False)  # -> ?active=is.false
client.table("t").select("*").is_("active", None)   # -> ?active=is.null   (unchanged)
```

## Additional context

The fix lives in the shared `base_request_builder.py`, so it covers both the sync and async clients. Added `test_is_true` / `test_is_false` / `test_is_none` to the postgrest filter-builder tests (`_async` and `_sync`). The full postgrest unit suite passes and `ruff` check + format are clean.

### DIFF
--- a/src/postgrest/src/postgrest/base_request_builder.py
+++ b/src/postgrest/src/postgrest/base_request_builder.py
@@ -366,6 +366,10 @@ class BaseFilterRequestBuilder(Generic[C]):
         """
         if value is None:
             value = "null"
+        elif value is True:
+            value = "true"
+        elif value is False:
+            value = "false"
         return self.filter(column, Filters.IS, value)
 
     def like(self: Self, column: str, pattern: str) -> Self:
--- a/src/postgrest/tests/_async/test_filter_request_builder.py
+++ b/src/postgrest/tests/_async/test_filter_request_builder.py
@@ -282,6 +282,26 @@ def test_is_(filter_request_builder):
     assert str(builder.request.params) == "x=is.a"
 
 
+def test_is_true(filter_request_builder):
+    # PostgREST's `is` operator is case-sensitive and only accepts lowercase
+    # trilean literals, so a Python ``True`` must become ``is.true``.
+    builder = filter_request_builder.is_("x", True)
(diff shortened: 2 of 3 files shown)


brandonchinn178:
FYI this is a general issue: https://github.com/supabase/supabase-py/issues/1583

winklemad:
Thanks for the pointer — #1583 is the right umbrella for the general `.eq()`/filter value-serialization behavior. I'd see this one as a narrower, orthogonal correctness bug though: `is` is a case-sensitive PostgREST operator that only accepts the lowercase trilean literals (`null`/`true`/`false`/`unknown`), so `is.True` is simply an invalid query today regardless of how the broader serialization design lands. Happy to fold it into a general serialization pass if you'd rather solve #1583 holistically — but since `is_()` already special-cases `None` right above this, mirroring that for booleans keeps it self-contained and safe to land now. Let me know which you'd prefer.

PR #1539, opened 2026-07-13
fix(utils): remove stray debug print in AsyncHttpIO

tushardev-365:
Removes a leftover debug print in AsyncHttpIO.communicate. Every async request on the v3 branch currently prints its full response to stdout. Noticed while reading the v3 work discussed in #1251.


tushardev-365:
@olirice @silentworks tagging per the contributing guide. One line change on the v3 branch, so also cc @o-santi.

o-santi:
Done in #1543, but thanks for noticing.

PR #1518, opened 2026-06-19
chore: wire up SDK compliance checking

grdsdev:
## Summary

Integrates supabase-py with the cross-SDK capability matrix at [supabase/sdk](https://github.com/supabase/sdk). On every push to `main` and every PR, CI will now:

1. Validate `sdk-compliance.yaml` against the canonical feature registry
2. Parse the Python public API surface and check that any new symbols added in a PR are declared in the compliance file

## Changes

- **`sdk-compliance.yaml`** — declares implementation status for all 195 features in the registry, with Python class/method symbols for cross-referencing (e.g. `AsyncGoTrueClient.sign_up`, `AsyncBucketActionsMixin.upload`)
- **`.sdk-parse-ignore`** — tells the Python parser to skip test directories and conftest files
- **`.github/workflows/validate-capabilities.yml`** — calls the reusable `validate-sdk-compliance.yml@main` workflow from supabase/sdk with `language: python`

## Implementation status highlights

| Area | Status | Notes |
|------|--------|-------|
| Auth (sign-in, session, identities, MFA, auto-refresh) | ✅ Implemented | |
| Auth admin (create/delete/update/get/list user, generate link, invite) | ✅ Implemented | |
| Auth admin MFA (list/delete factor) | ✅ Implemented | Exposed via `admin.mfa.*` |
| OAuth admin (create/delete/get/list/update client, regenerate secret) | ✅ Implemented | Exposed via `admin.oauth.*` |
| Database (CRUD, filters, modifiers, explain, or_, select-after-mutation) | ✅ Implemented | `select()` chaining after insert/update/upsert/delete |
| Database filters (not_in, like_all_of, like_any_of, ilike_all_of, ilike_any_of) | ✅ Implemented | `not_in` via `.not_.in_()`; rest via `BaseFilterRequestBuilder` |
| Database modifiers (max_affected_rows, request_cancellation, request_timeout) | ✅ Implemented | Cancellation via asyncio task `.cancel()`; timeout via `asyncio.timeout()` |
| Storage (file buckets, signed URLs) | ✅ Implemented | |
| Storage vector buckets (create/list/delete bucket, indexes, put/get/list/query/delete vectors, parallel scan) | ✅ Implemented | Via `storage.vectors()` → `AsyncStorageVectorsClient` |
| Storage analytics buckets (create/list/delete, Iceberg namespace/table) | ✅ Implemented | Via `storage.analytics()` → `AsyncStorageAnalyticsClient` + pyiceberg `RestCatalog` |
[...]

PR #1540, opened 2026-07-14
fix(ci): point SDK compliance workflow at the python reusable workflow

tushardev-365:
`validate-capabilities.yml` calls `supabase/sdk/.github/workflows/validate-sdk-compliance.yml@main`, but that workflow does not exist in `supabase/sdk`. The compliance workflows there are per language: `validate-sdk-compliance-python.yml`, `-dart`, `-javascript`, `-swift`.

So the job fails before it runs on every PR:

```
error parsing called workflow ".github/workflows/validate-capabilities.yml"
-> "supabase/sdk/.github/workflows/validate-sdk-compliance.yml@main"
: failed to fetch workflow: workflow was not found.
```

This is happening on all open PRs right now, including dependabot's and internal ones, not just fork PRs.

This points the job at `validate-sdk-compliance-python.yml` and drops the `language` input, which that workflow does not declare (the language is implied by which workflow you call). `griffe-packages` and `griffe-search-paths` are unchanged and match the inputs it expects.

Introduced in #1518.

### DIFF
--- a/.github/workflows/validate-capabilities.yml
+++ b/.github/workflows/validate-capabilities.yml
@@ -11,9 +11,8 @@ permissions:
 
 jobs:
   validate:
-    uses: supabase/sdk/.github/workflows/validate-sdk-compliance.yml@main
+    uses: supabase/sdk/.github/workflows/validate-sdk-compliance-python.yml@main
     with:
-      language: python
       griffe-packages: >-
         supabase supabase_auth postgrest storage3 realtime supabase_functions
       griffe-search-paths: >-


tushardev-365:
@grdsdev tagging you since this touches the compliance workflow from #1518, and @silentworks per the contributing guide. Happy to adjust if the intent was a shared dispatcher workflow in supabase/sdk rather than the per-language ones.

grdsdev (approved):
Thanks!

PR #1543, opened 2026-07-17
feat(realtime): realtime major overhaul

o-santi:

Rewrite realtime from scratch, aiming at removing most callbacks from the core library, by instead offering an event iterator based library, upon which callbacks can more safely be built upon.

The current library suffers from 3 problems that hinder usability: 
1. There is no simple way to know which events arrived happened first, as the callbacks can be called at any time. When the user wants to ensure some particular order he needs to shove a locking mechanism into the callback and wait for it on the outside. This can get quite complicated.
2. All exceptions that happen on background tasks are ignored. This includes exceptions thrown from callbacks, as well as from parsing and sending messages in behalf of the user.
3. Automatic re connection logic happens in background threads with no stack to bubble up errors. This can cause mysterious behavior if the background connection reaches max_retries and silently dies after hours or days of normal use. Automatic reconnection in the background also has weird semantic problems, like not knowing if a message may have been sent twice, or 

This PR proposes to get rid of all of these behaviors, by doing as much as possible outside of background threads. I believe this can be mostly solved by offering fewer and simpler building blocks using iterators; ie. instead of
```python
channel.on_broadcast(my_broadcast_callback)
channel.on_presence(my_presence_callback)
subscribe_event = asyncio.Event()
channel.on_subscribe(lambda _: subscribe.set())

await subscribe_event.wait()

await channel.send_broadcast("my-event", "i'm subscribed")
```
I propose doing:
```python
subscribe_ack = await channel.subscribe()

send_broadcast_ack = await channel.send_broadcast("my-event", "i'm subscribed")

async for message in channel.messages():
    if isinstance(message, BroadcastMessage):
         my_broadcast_callback(message)
    elif isinstance(message, PresenceMessage):
         my_presence_callback(message)
```

which is both easier to understand and has explicit ordering between events

# Core design decisions

## Listen and heartbeat still happen in the background

The idea is to offer per-channel message listeners, but internally we only hold one websocket connection. This means that the listen task needs to happen in the background in order to multiplex the messages and route them to the correct channel object. Because of this, we still need a way to bubble up errors in these background tasks. 

The proposed solution is to have a signaling `asyncio.Future` called `self._connection_failure`, that signals if an error happened in either the `listen` or the `heartbeat` task. Then, each message that reads from the websocket connection (be it waiting for an ack message or listening for messages on a channel) must poll both futures (the receive one and the failure future) at once, in order to bubble up the exception if it happens.

## Exceptions may bubble up on all await points

In order to not let connection failures be silently ignored, the library intends to raises any connection errors (in fact, any un-checked exception that may happen at all) to the user. As it may happen in the background at any point, it needs to have a stack where it can raise them. This means that, if the user is not currently polling the library, errors may already have happened and will be thrown the next time he tries to interact with it, be it by listening to messages or by trying to send one.

This also means that re-connection now needs to be handled with more care, as it is not being handled implicitly anymore; which brings up the next point:

## Automatic re connection is not entirely solved

I'm not sure that the current behavior of automatically reconnecting when a connection occurs is the best behavior to have as default in the library. The problem here is largely that, when an error occurs and the client automatically reconnects, it is not in the same state as before: all channels are gone.

The way the library works today is to pretend that nothing happened and try to reconnect in the background to all channels were marked as joined before the disconnect occurred, and continue from where it left off. This has some weird edge cases, where some messages may be dropped entirely if the connection error was raised inside a `send_*` function. Furthermore, the more pressing issue is that it pretends that all channel connections are entirely stateless, which might not be the case for all applications.

If a user wants to ensure that a special method is called or a particular message is sent whenever a user joins or leaves a channel, it is not possible today. This is because there is no way to know from the outside when a channel is being rejoined in the background, or was left due to a connection error. All of this implies that trying to build a stateful workflow on top of the current library is very hard, and I'm not sure there is a way to reconcile this.

`automatically_reconnect` is **a** way to try to solve this, by using a blocking function that expects a callback, and calls it with the client whenever a connection failure happened and it was able to reconnect. The intended use cases is the user knowing when a re-connection occured, by the callback being called again, and having tools for dealing with it in his own way; for instance, he may choose to track the current open channels and re-connect as if nothing happend, if the old behavior was okay for him.

## A callback based design is still possible

The idea with offering simpler building blocks is that bulding callback based abstractions is still possible; in fact it should be a lot easier now, as it should only worry about calling the callbacks and dealing with the acks, all the messaging parsing and sending logic should be dealt with already, and the user may treat each channel as it's own connection entirely.

In fact, the callback based design is so straightforward to write that I'm unsure if we should offer it as one of the building blocks of the library, or let the user write the one he wants. It should be as simple as:
```python
class ChannelCallbackManager:
    async def run(self):
        async for message in self.channel.messages():
            if isinstance(message, BroadcastMessage):
                for callback in self.broadcast_callback:
                    await callback(message)
            ...

    def on_broadcast(self, callback: ...):
        self.broadcast_callbacks.append(callback)
```


src/realtime/src/realtime/channel.py
```diff
@@ -0,0 +1,319 @@
...
+            join_ref=None,
+        )
+        msg = await self.socket.send(message)
+        logger.info(f"Subscribe reply: {msg!r}")
+        self.join_ref = self.socket._make_ref()
+        if msg.payload.status != "ok":
+            raise Exception(
+                f"error while subscribing to channel: {msg.payload.response!r}"
+            )
+        return msg
```
grdsdev, line 217: `subscribe()` never marks the channel joined → `set_auth()` is a no-op, `subscribe()` sends the join, checks the reply, returns — but never sets `self.joined = True` or `self.state = ChannelStates.JOINED`. They stay at the `__init__` defaults (False / CLOSED).

Consequences:

- `client.set_auth()` (client.py) filters on `channel.joined` and `channel.state == ChannelStates.JOINED` → always False → token refresh is never pushed to any channel.

- `_can_push()` (uses self.joined) is always False.
Fix: set `self.joined = True / self.state = JOINED` on a successful subscribe reply (and reset on unsubscribe/error).

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+        try:
+            while True:
+                msg = await self._ws_connection.recv(decode=False)
+                try:
+                    message = Message.model_validate_json(msg)
+                except ValidationError as e:
+                    logger.error(f"Unrecognized message format {msg!r}\n{e}")
+                    continue
+                if (message.ref is not None) and self.ack:
+                    self.message_refs[message.ref].set_result(message)
```
grdsdev, line 181: `KeyError` if key is absent, should catch this?

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+) -> None:
+    retries = 0
+    while True:
+        try:
+            async with connect_once(url, token) as client:
+                retries = 0
+                await client_handler(client)
+                break
+        except websockets.ConnectionClosed:
+            continue
```
grdsdev, line 133: No delay here?

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+        self._ws_connection: ClientConnection = ws_connection
+        self.ref = 0
+        self.channels: Dict[str, RealtimeChannel] = {}
+        self.message_refs: Dict[str, asyncio.Future[Message]] = {}
+        self._connection_failure: asyncio.Future[Message] = asyncio.Future()
+        self.heartbeat_timeout = heartbeat_timeout
+
+    @property
+    def is_connected(self) -> bool:
+        return self._ws_connection is not None
```
grdsdev: `self._ws_connection` is never cleaned, so this is always True.

src/realtime/src/realtime/channel.py
```diff
@@ -0,0 +1,319 @@
...
+        message = Message(
+            topic=self.topic,
+            event=ChannelEvents.join,
+            payload=payload.model_dump(exclude_none=True),
+            ref=self.socket._make_ref(),
+            join_ref=None,
+        )
+        msg = await self.socket.send(message)
+        logger.info(f"Subscribe reply: {msg!r}")
+        self.join_ref = self.socket._make_ref()
```
grdsdev: `self.join_ref` only exists after `subscribe()`, consider assigning `None` to it on `__init__`.

src/realtime/src/realtime/channel.py
```diff
@@ -0,0 +1,319 @@
...
+    Each Channel has its own topic and a list of event-callbacks that respond to messages.
+    Should only be instantiated through `AsyncRealtimeClient.channel(topic)`.
+    """
+
+    def __init__(
+        self,
+        socket: RealtimeClient,
+        topic: str,
+        params: RealtimeChannelOptions | None = None,
+        ack: bool = True,
```
grdsdev: What's this `ack`?

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+        return url.with_scheme("wss")
+    return url.with_scheme("ws")
+
+
+class connect_once:
+    def __init__(
+        self,
+        url: str,
+        token: str | None = None,
+        params: Dict[str, Any] | None = None,
```
grdsdev: never used.

src/realtime/pyproject.toml
```diff
@@ -12,11 +12,13 @@ maintainers = [
 license = "MIT"
 readme = "README.md"
 repository = "https://github.com/supabase/supabase-py"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
   "websockets >=11,<16",
   "typing-extensions >=4.14.0",
   "pydantic (>=2.11.7,<3.0.0)",
+  "pytest-asyncio>=0.21.0",
```
grdsdev: should this be added to dev dependencies instead?

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+
+        message_str = message.model_dump_json()
+        logger.debug(f"sending: {message_str}")
+
+        receive: asyncio.Future[Message] = asyncio.Future()
+        self.message_refs[message.ref] = receive
+        await self._ws_connection.send(message_str)
+        logger.debug(f"waiting for ref='{message.ref}'")
+
+        response = await self.get_message_or_raise(receive)
```
grdsdev, line 332: if this call raises then `del self.message_refs[message.ref]` is not called.

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+                raise MaxRetriesExceeded(retries)
+
+
+class RealtimeClient:
+    def __init__(
+        self,
+        url: URL,
+        ws_connection: ClientConnection,
+        token: str | None = None,
+        hb_interval: int = DEFAULT_HEARTBEAT_INTERVAL,
```
grdsdev, line 149:
```suggestion
        heartbeat_interval: int = DEFAULT_HEARTBEAT_INTERVAL,
```

src/realtime/src/realtime/client.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+def exponential_delay(max_retries: int, retry_count: int) -> float | None:
+    if retry_count == max_retries:
+        return None
+    else:
+        return float(2**retry_count)
+
+
+async def automatically_reconnect(
```
grdsdev, line 119: we had discussed about naming this `connect`, I'm not used to use Python, but it seems calling `connect` is the right thing, considering the `websocket` also names it connect.

src/realtime/src/realtime/channel.py
```diff
@@ -0,0 +1,319 @@
...
+    token: str | None = None
+    postgres_changes_bindings: list[PostgresChangesBinding] = field(
+        default_factory=list
+    )
+
+    def broadcast(
+        self,
+        ack: bool = False,
+        listen_self: bool = False,
+        replay: float | None = None,
```
grdsdev: what is this `replay`? it is not used.

grdsdev (changes requested):

src/realtime/src/realtime/channel.py
o-santi: `set_auth` is not a no-op because it sets the `access_token` value, that's why the tests have been working so far. But indeed I forgot to update that variable.

src/realtime/src/realtime/client.py
o-santi: True, this property is not used anywhere anymore, so I should just remove it.

src/realtime/src/realtime/client.py
o-santi: Removed.

src/realtime/src/realtime/channel.py
o-santi: A flag for disabling the `ack` awaiting, but I don't think this is a great idea, as it effectively makes 2 clients with 2 different behaviors (and possibily different bugs). I just removed it.

src/realtime/pyproject.toml
o-santi: Yes, we already have it in our dev-dependencies, I'm not sure how this got here.

src/realtime/src/realtime/client.py
o-santi: This is a small optimization to avoid having long running objects hogging a lot of memory due to all message's acks being referenced in this dict. If this call raises, then the object is going to be GC'd eventually, as the user is instructed to create a new one when connection errors happen, so we don't really need to delete the object.

src/realtime/src/realtime/channel.py
o-santi: A argument that was renamed and I forgot to remove. Thanks for noticing.

src/realtime/src/realtime/client.py
o-santi: The difference is that `websockets` does not differentiate between these two cases. Bare `async with connect() ...` is `connect_once`, and in order to "automatically reconnect" you need to iterate the connection and catch connection errors yourself:
```python

async for websocket in connect(...):
    try:
        do_stuff(websocket)
    except ConnectionClosed:
        continue
```
`automatically_reconnect` is a wrapper over this pattern with exponential retries when it disconnects, where you pass in the `do_stuff` callback. I do not think it makes sense to call either of these `connect`, for it might give the user the sense that it is doing background connection, or that one is "preferred" over the other.

src/realtime/src/realtime/client.py
o-santi: If this raises `message.ref` then we either have a bug in the library or in the server, and in both cases we'd like to know. 

A core property of `ref`s is that they always refer to some message that we have sent previously, and have thus registered in the `message_refs` dict; if that is not the case, then the library or the server certainly has a bug.

src/realtime/src/realtime/client.py
o-santi: This case should only occur when you had an established connection and the `client_handler` function raises a connection issue. This exception does not necessarily imply a network error; so I'm not sure it makes sense to have a delay here. If the immediate next connection fails, it falls in the except case bellow, where it will have the normal delay.

In my understanding, exponential delays exist when the network falters and you cannot reach the server for some unknown time; and I'm not sure that it is the case here.

grdsdev:
## Realtime capability matrix check

*Generated by Claude, checked against [supabase/sdk `capabilities/realtime.yaml`](https://github.com/supabase/sdk/blob/main/capabilities/realtime.yaml) (27 capabilities) by inspecting the actual code on this branch (not just the diff).*

**Tally: 18 Implemented, 1 Partial, 2 Removed/Regressed, 6 Not Implemented**

### client
| Capability | Status | Evidence |
|---|---|---|
| `client.channel` | ✅ Implemented | `client.py:241-254` |
| `client.connect` | ✅ Implemented | `client.py:56-90` (moved to `connect_once` context manager) |
| `client.connection_state` | 🔴 Removed/Regressed | No `is_connected`/state accessor remains; the old client exposed one |
| `client.disconnect` | ✅ Implemented | `client.py:193-204`, renamed to `close()` |
| `client.get_channels` | ✅ Implemented | `client.py:256-257` |
| `client.listen_heartbeats` | ⚪ Not Implemented | Heartbeat only logs internally, no callback hook (gap pre-existed) |
| `client.remove_all_channels` | 🟡 Partially Implemented | Unsubscribes each channel but never clears `self.channels` dict (`client.py:267-275`) |
| `client.remove_channel` | ✅ Implemented | `client.py:262-265` |
| `client.set_auth_token` | ✅ Implemented | `client.py:277-315`, renamed to `set_auth()` |

### channel
| Capability | Status | Evidence |
|---|---|---|
| `channel.broadcast_http` | ⚪ Not Implemented | URL builder exists (`channel.py:185`) but never wired to an HTTP call — dead code carried over from the old lib, not a new regression |
| `channel.send` | ✅ Implemented | `send_broadcast()`, `channel.py:289-301` |
| `channel.subscribe` | ✅ Implemented | `channel.py:198-217` |
| `channel.unsubscribe` | ✅ Implemented | `channel.py:238-239` |

### subscriptions
| Capability | Status | Evidence |
|---|---|---|
| `subscriptions.broadcast` | ✅ Implemented | `BroadcastMessage` yielded via `channel.messages()` iterator |
| `subscriptions.postgres_changes` | ✅ Implemented | `channel.py:103-115` |
| `subscriptions.subscribe_presence` | ✅ Implemented | `PresenceJoin`/`PresenceLeave` via the iterator |
| `subscriptions.postgres_changes_filter` | ✅ Implemented | |
| `subscriptions.postgres_changes_multiple_filters` | ✅ Implemented | bindings list, chained `.postgres_changes()` calls |
| `subscriptions.private_channel` | ✅ Implemented | `.private(True)` option |
| `subscriptions.broadcast_self` | ✅ Implemented | `broadcast(listen_self=...)` |
| `subscriptions.broadcast_ack` | ✅ Implemented | `channel.py:297-301` |
| `subscriptions.broadcast_replay` | ✅ Implemented | `replay_since`/`replay_limit`, covered by tests |

### presence
| Capability | Status | Evidence |
|---|---|---|
| `presence.presence_state` | ✅ Implemented | `channel.py:256-262` |
| `presence.track` | ✅ Implemented | |
| `presence.untrack` | ✅ Implemented | |
| `presence.presence_key` | ✅ Implemented | `presence(key=...)` |

### configuration
| Capability | Status | Evidence |
|---|---|---|
| `configuration.custom_websocket_transport` | ⚪ Not Implemented | `websockets.connect()` hardcoded, no injection point (pre-existing gap) |
| `configuration.reconnect_backoff` | 🔴 Removed/Regressed | Old client had an `initial_backoff` param; new `exponential_delay()` is fully hardcoded, no override |
| `configuration.heartbeat_interval` | ✅ Implemented | `hb_interval` param threads through |
| `configuration.access_token_callback` | ✅ Implemented | `token_callback`, refreshed on every heartbeat |
| `configuration.deferred_disconnect` | ⚪ Not Implemented | Never present, old or new |
| `configuration.custom_logger` | ⚪ Not Implemented | Module-level logger, no injection point |
| `configuration.binary_protocol` | ⚪ Not Implemented | `VSN` hardcoded to `"1.0.0"`, JSON only |

### Summary

Two real regressions introduced by this rewrite: **`connection_state`** and **`reconnect_backoff`** customization both existed before and are gone now. The other five gaps (`broadcast_http`, `custom_websocket_transport`, `deferred_disconnect`, `custom_logger`, `binary_protocol`) predate this PR — they're still-open gaps against the matrix, not new regressions from this change.

PR #1557, opened 2026-07-31
feat(postgrest): support single()/maybe_single() after insert/upsert select

### LINKED ISSUE #1553: Allow selecting single row after INSERT
### Description

I believe the JS client supports
```js
client.from("table").insert(...).select().single()
```
This doesn't work in the Python client, since `.insert()` returns `SyncQueryRequestBuilder` and `.select()` returns `self`.

### Suggested solution

The JS client returns the same `PostgrestFilterBuilder` for select/insert. Maybe that should be done here

### Alternatives

_No response_

### Additional context

_No response_

tushardev-365:
## What

Closes #1553.

The JS client supports `client.from("t").insert(...).select().single()`, but the Python client didn't: `insert()` and `upsert()` return a `*QueryRequestBuilder`, whose `select()` returns `self`, and `single()` / `maybe_single()` only lived on `*SelectRequestBuilder`. So after an insert/upsert there was no way to get a single-row response.

## Change

Move `single()` and `maybe_single()` up from `*SelectRequestBuilder` to the base `*QueryRequestBuilder`. They're now reachable from every builder `select()` can return, so all of these work and mirror the JS client:

```py
client.from_("t").insert(row).select().single().execute()
client.from_("t").upsert(row).select().maybe_single().execute()
```

`*SelectRequestBuilder` still exposes them via inheritance, so existing `select().single()` usage is unchanged. The methods only set headers and return the existing single/maybe-single builders, so there's no behavioural change to what `single()` itself does.

Async is the source of truth; the sync builder and both test files are the unasync-generated mirror.

## Testing

Added `test_select_single` / `test_select_maybe_single` to the query-request-builder tests (async + sync), asserting `select().single()` returns the single-request builder with the `application/vnd.pgrst.object+json` Accept header and `select().maybe_single()` returns the maybe-single builder. Full postgrest unit suite passes:

```
90 passed
```

### DIFF
--- a/src/postgrest/src/postgrest/_async/request_builder.py
+++ b/src/postgrest/src/postgrest/_async/request_builder.py
@@ -75,6 +75,19 @@ class AsyncQueryRequestBuilder:
         self.request.retry_enabled = enabled
         return self
 
+    def single(self) -> AsyncSingleRequestBuilder:
+        """Specify that the query will only return a single row in response.
+
+        .. caution::
+            The API will raise an error if the query returned more than one row.
+        """
+        self.request.headers["Accept"] = "application/vnd.pgrst.object+json"
+        return AsyncSingleRequestBuilder(self.request)
+
+    def maybe_single(self) -> AsyncMaybeSingleRequestBuilder:
+        """Retrieves at most one row from the result. Result must be at most one row (e.g. using `eq` on a UNIQUE column), otherwise this will result in an error."""
+        return AsyncMaybeSingleRequestBuilder(self.request)
+
     async def execute(self) -> APIResponse:
         """Execute the query.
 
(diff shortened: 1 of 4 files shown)


o-santi:
I'm afraid that this might allow `.single()` for queries that do not make sense to do it. Is this how the JS library is structured?

tushardev-365:
Yeah, this matches how JS is structured. There `insert`/`upsert`/`update`/`delete`/`select` all return `PostgrestFilterBuilder`, and `single()`/`maybeSingle()` live on its parent `PostgrestTransformBuilder`, so they're reachable after all of those, same as this PR.

JS doesn't gate the "doesn't make sense" cases either: `single()` just sets the `Accept: application/vnd.pgrst.object+json` header and PostgREST decides the result. So this brings us in line with the JS client rather than exposing anything beyond it. Happy to scope it tighter if you'd prefer, though.

PR #1567, opened 2026-08-08
fix(storage): surface real error when error body lacks message

### LINKED ISSUE #1563: storage3: AttributeError masks the real error when a storage error body lacks message/error/statusCode
## Bug report

`SyncBucketProxy._request` (and its async twin) raises `AttributeError: 'dict' object has no attribute 'text'` instead of a `StorageApiError` whenever the storage API returns a non-2xx whose JSON body is missing any of `message`, `error` or `statusCode`.

The real status and body are discarded, so the caller learns nothing about what actually failed.

## Where

`src/storage/src/storage3/_sync/file_api.py`, in `_request`:

```python
    try:
        resp = exc.response.json()
        raise StorageApiError(
            resp["message"], resp["error"], resp["statusCode"]
        ) from exc
    except KeyError as err:
        message = f"Unable to parse error message: {resp.text}"
        raise StorageApiError(message, "InternalError", 400) from err
```

`resp` is the parsed **dict** from `exc.response.json()`, so `resp.text` on the recovery path always raises. The attribute being reached for belongs to the response, not to the decoded body — `exc.response.text`.

The same code is in `src/storage/src/storage3/_async/file_api.py`.

## Reproduction

```python
import httpx
from storage3._sync.file_api import SyncBucketProxy

# Any storage error whose body lacks one of message/error/statusCode, e.g.
# {"code": "...", "msg": "..."} from a gateway or a non-storage layer.
```

Minimal equivalent of the failing branch:

```python
resp = {"code": "TooLarge"}          # no "message" key
resp["message"]                       # KeyError -> recovery path
f"{resp.text}"                        # AttributeError: 'dict' object has no attribute 'text'
```

Expected: `StorageApiError("Unable to parse error message: {\"code\": \"TooLarge\"}", "InternalError", 400)`
Actual: `AttributeError: 'dict' object has no attribute 'text'`, with the original `HTTPStatusError` only reachable via `__context__`.

## Suggested fix

```python
    except KeyError as err:
        message = f"Unable to parse error message: {exc.response.text}"
        raise StorageApiError(message, "InternalError", 400) from err
```

Happy to open a PR for both the sync and async copies if that's useful.

## Impact

This is the error path, so it only fires when something is already wrong — which is exactly when the message matters. In our case it masked a reverse-proxy timeout severing large uploads: every occurrence reached our logs as `'dict' object has no attribute 'text'` and nothing else, and diagnosing it meant reading the storage service's own access logs instead. We now unwrap `__context__` to recover the status and body, which is not something callers should have to do.

Version: `storage3` 2.31.0, and the code is unchanged on `main` at the time of writing.

MRoshaan:
## Description
In `storage3`, when the Storage API returns a non-2xx response whose JSON body is missing any of `message`, `error` or `statusCode`, the error-recovery path accessed `resp.text` where `resp` is the already-decoded `dict` — raising `AttributeError: 'dict' object has no attribute 'text'` and completely masking the real error.

## Motivation
Closes #1563. The real status and body were being discarded exactly when they matter most (the error path). The fix reports the raw response text in a proper `StorageApiError` instead.

## Changes
- `src/storage3/_sync/file_api.py` and `src/storage3/_async/file_api.py`: use `exc.response.text` on the recovery path.
- Added regression tests (`tests/_sync/test_file_api.py`, `tests/_async/test_file_api.py`) covering both sync and async `_request`.

## Testing
- New tests pass for `_sync` and `_async`.
- Adjacent storage unit tests (bucket, exceptions, utils) pass: 28 passed.
- `ruff check` and `ruff format --check` clean on changed files.

Cc: cc @olirice @silentworks for review.

### DIFF
--- a/src/storage/src/storage3/_async/file_api.py
+++ b/src/storage/src/storage3/_async/file_api.py
@@ -84,7 +84,7 @@ class AsyncBucketActionsMixin:
                     resp["message"], resp["error"], resp["statusCode"]
                 ) from exc
             except KeyError as err:
-                message = f"Unable to parse error message: {resp.text}"
+                message = f"Unable to parse error message: {exc.response.text}"
                 raise StorageApiError(message, "InternalError", 400) from err
 
         # close the resource before returning the response
--- a/src/storage/src/storage3/_sync/file_api.py
+++ b/src/storage/src/storage3/_sync/file_api.py
@@ -84,7 +84,7 @@ class SyncBucketActionsMixin:
                     resp["message"], resp["error"], resp["statusCode"]
                 ) from exc
             except KeyError as err:
-                message = f"Unable to parse error message: {resp.text}"
+                message = f"Unable to parse error message: {exc.response.text}"
                 raise StorageApiError(message, "InternalError", 400) from err
 
         # close the resource before returning the response
--- a/src/storage/tests/_async/test_file_api.py
+++ b/src/storage/tests/_async/test_file_api.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+import httpx
+import pytest
+from httpx import Headers
+from storage3._async.file_api import AsyncBucketProxy
+from storage3.exceptions import StorageApiError
+from yarl import URL
+
+
+async def test_request_raises_real_error_when_body_missing_message_field() -> None:
+    """A non-2xx body without message/error/statusCode must surface as StorageApiError."""
+
+    def handler(request: httpx.Request) -> httpx.Response:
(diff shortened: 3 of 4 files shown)


grdsdev:
Thanks for this, and for adding tests — that put it ahead of the other one-line fixes for this bug.

Going with #1577 instead, because it also covers `bucket.py::_request`, which has the same bug in a worse form (no `except KeyError` at all, so a missing key raises a bare `KeyError`), and it catches `ValueError` so non-JSON error bodies are handled rather than raising `JSONDecodeError` before the handler is reached. Sorry for the slow turnaround here — there were seven PRs open against this one issue.

PR #1574, opened 2026-08-15
fix(storage): raise StorageApiError when a bucket error body is unrecognized

tushardev-365:
## What

`StorageBucketAPI._request` indexed the parsed error body directly:

```python
    resp = exc.response.json()
    raise StorageApiError(resp["message"], resp["error"], resp["statusCode"]) from exc
```

So a storage error whose JSON body is missing any of `message` / `error` / `statusCode` came back as a bare `KeyError`, and a body that isn't JSON at all (an HTML page from a proxy or gateway) came back as a `JSONDecodeError`. Either way the caller lost the real status and body and got an opaque exception instead of `StorageApiError` — on the one path where the message matters most. This affects every bucket operation: `list_buckets`, `get_bucket`, `create_bucket`, `update_bucket`, `empty_bucket`, `delete_bucket`.

## Fix

Wrap the lookup and fall back to `StorageApiError` carrying the raw response text, mirroring the shape `file_api._request` already uses and the `ValidationError` fallback `request.py` uses for vector buckets.

## Tests

Two new cases per flavour (async + sync), asserting `StorageApiError` rather than the leaked exception:
- error body missing the expected keys
- error body that isn't JSON

Reverting only the source change makes all four fail with the raw `KeyError` / `JSONDecodeError`; with the fix the bucket suite is 28 passed.

## Note

#1563 reports this same class of bug in `file_api._request` (its recovery path reaches for `resp.text` on a dict), and there are already PRs open for that file. This PR is the bucket API only, which is a separate `_request` with no recovery path at all — so the two don't overlap.

`_sync` is generated, so the change was made in `_async` and regenerated with `make build-sync`; only the bucket files are included here.

### DIFF
--- a/src/storage/src/storage3/_async/bucket.py
+++ b/src/storage/src/storage3/_async/bucket.py
@@ -42,10 +42,14 @@ class AsyncStorageBucketAPI:
             )
             response.raise_for_status()
         except HTTPStatusError as exc:
-            resp = exc.response.json()
-            raise StorageApiError(
-                resp["message"], resp["error"], resp["statusCode"]
-            ) from exc
+            try:
+                resp = exc.response.json()
+                raise StorageApiError(
+                    resp["message"], resp["error"], resp["statusCode"]
+                ) from exc
+            except (KeyError, ValueError) as err:
+                message = f"Unable to parse error message: {exc.response.text}"
+                raise StorageApiError(message, "InternalError", 400) from err
 
         return response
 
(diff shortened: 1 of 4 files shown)


grdsdev:
This was the strongest of the two-file analyses — you were the only one besides #1577 to spot that `exc.response.json()` raising `ValueError` bypasses the `KeyError` handler entirely, and your non-JSON-body test covers exactly the gateway-timeout case the reporter hit.

Closing in favor of #1577, which applies that same `(KeyError, ValueError)` fix to both `bucket.py` and `file_api.py` in one pass, plus keeps the real status code. Your read on the bug was right; it just landed as a subset of the wider patch.

PR #1577, opened 2026-08-16
fix: handle non-standard and non-JSON error responses without AttributeError

Vishv07:
### Summary

Fixes an `AttributeError` in `storage3` where non-2xx responses that lack `message`, `error`, or `statusCode` keys in their JSON body (or return non-JSON payloads like HTML 502/504 gateway errors) crashed with `'dict' object has no attribute 'text'` or uncaught `JSONDecodeError`.

### Changes

- In `_async/file_api.py`, `_async/bucket.py`, and `_sync` counterparts:
  - Catch `(KeyError, ValueError)` when decoding error response JSON.
  - Access `exc.response.text` instead of `resp.text`.
  - Pass `exc.response.status_code` to `StorageApiError`.
- Added unit regression tests for missing error keys and non-JSON error responses across `test_client.py` and `test_bucket.py` (both async and sync).

### Verification

- All 51 unit tests in `src/storage` pass.
- `mypy` check passes with 0 issues.
- `ruff check` and `ruff format` pass cleanly.

### DIFF
--- a/src/storage/src/storage3/_async/bucket.py
+++ b/src/storage/src/storage3/_async/bucket.py
@@ -42,10 +42,16 @@ class AsyncStorageBucketAPI:
             )
             response.raise_for_status()
         except HTTPStatusError as exc:
-            resp = exc.response.json()
-            raise StorageApiError(
-                resp["message"], resp["error"], resp["statusCode"]
-            ) from exc
+            try:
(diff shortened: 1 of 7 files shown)


Vishv07:
@o-santi can you please review the PR?

grdsdev (approved):

Issue #1585, opened 2026-08-20
count with head=True fails when filtering on embedded resource


brandonchinn178:
### Describe the bug

I want to get the count of a query filtering on an embedded resource using `head=True`, which fails because postgrest-py strips out select on head requests:

```
'ref' is not an embedded resource in this request
```

### Reproduction

```python
`client.from_("my_table").select("*, ref!inner(*)", count=CountMethod.exact, head=True).eq("ref.foo", 123).execute()`
```

### Library affected

postgrest

### Library version

main

### Python version

3.13

bindu-yogesh:
Hi! I'd like to work on this issue. I've gone through the reproduction and would like to investigate the head=True / embedded-resource behavior and add a regression test. Could you please assign this to me?

PR #1588, opened 2026-08-24
test(postgrest)- Cover HEAD count with embedded filters

Pranava-Pai-N:
Bug fix regression test for PostgREST HEAD requests with embedded-resource filters.

Queries using `head=True`, an exact count, and an embedded-resource filter can fail when the `select` parameter is missing:

```text
'ref' is not an embedded resource in this request
```

#1585 

## What is the new behavior?
- Add sync and async regression tests confirming that HEAD requests retain:
- The embedded-resource select expression
- The embedded-resource filter
- The count=exact preference
- The HEAD HTTP method

## Additional context
The current implementation already preserves the select parameter for HEAD requests. This PR adds regression coverage to prevent that behavior from breaking in the future.

### DIFF
--- a/src/postgrest/tests/_async/test_request_builder.py
+++ b/src/postgrest/tests/_async/test_request_builder.py
@@ -45,6 +45,20 @@ class TestSelect:
         assert builder.request.http_method == "HEAD"
         assert builder.request.json is None
 
+    def test_select_with_head_and_embedded_filter(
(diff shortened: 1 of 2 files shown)


Pranava-Pai-N:
@o-santi Could you please review this PR solves issue #1585 and suggest any changes

Pranava-Pai-N:
@o-santi Any updates regarding the PR or any changes required

PR #1591, opened 2026-08-25
fix(postgrest): serialize filter values in eq()

### LINKED ISSUE #1583: .eq() needs to properly serialize values
### Describe the bug

Currently, `.eq()` just stringifies the Python value into the `eq.` param:
https://github.com/supabase/supabase-py/blob/80fcba7fde5c2347fa5327c937b4551754732ccb/src/postgrest/src/postgrest/base_request_builder.py#L302

This has surprising behavior when passing `None` or JSON values (for equality with json columns):
```
?col=eq.None
?col=eq.{'key':%20'value'}
```

Changes:
* `None` should serialize to `null` (will always return `False`, but it will run the correct SQL now)
* dicts and lists should JSON serialize
    * Alternatively, force users to explicitly serialize
* Error if value could not be serialized
* Probably affects other operators as well

Related: https://github.com/supabase/supabase-py/pull/1535 https://github.com/supabase/postgrest-py/pull/446

### Reproduction

```python
result = (
    client.from_("my_table")
    .select("*")
    .eq("my_col", None)
    .eq("json_col", {"key": "value"})
    .execute()
)
```

### Steps to reproduce

_No response_

### Library affected

postgrest

### Library version

main

### Python version

3.13

chengwudi1:
## What                                                                                                               
  Fixes #1583 — `.eq()` stringified filter values directly, so `None` produced `eq.None` (invalid SQL) and dict/list    
  values (for JSON columns) were serialized as Python reprs.                                                            
                                                                                                                        
  ## How                                                                                                                
  Added `_serialize_filter_value()` in `base_request_builder.py` and used it in `.eq()`:                                
  - `None` → `null`                                                                                                     
  - `bool` → `true` / `false`                                                                                           
  - `dict` / `list` → compact JSON (`{"key":"value"}`)                                                                  
  - everything else → unchanged `str()`                                                                                 
                                                                                                                        
  ## Tests                                                                                                              
  Added parametrized `test_eq_serializes_values` in both `tests/_async` and `tests/_sync` (7 cases: None, True, False,  
  dict, list, str, int).                                                                                                
                                                                                                                        
  Verified: reverting only the source change fails 5/7 cases. All 192 unit tests pass; `ruff check` and `ruff format`   
  clean.

### DIFF
--- a/src/postgrest/src/postgrest/base_request_builder.py
+++ b/src/postgrest/src/postgrest/base_request_builder.py
@@ -110,6 +110,21 @@ def _unique_columns(json: List[Dict[str, JSON]]):
     return columns
 
 
+def _serialize_filter_value(value: Any) -> str:
+    """Serialize a filter value into the wire format PostgREST expects.
+
+    ``None`` maps to ``null``, booleans to ``true``/``false``, and dicts/lists
+    to compact JSON, instead of the plain ``str()`` result (e.g. ``"None"``).
+    """
+    if value is None:
+        return "null"
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, (dict, list)):
+        return json.dumps(value, separators=(",", ":"))
+    return str(value)
+
+
 def _cleaned_columns(columns: Tuple[str, ...]) -> str:
     quoted = False
     cleaned = []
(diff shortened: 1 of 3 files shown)


chengwudi1:
Hi @grdsdev! Friendly ping on this one — it's been two weeks with no activity. The PR adds `_serialize_filter_value` so `eq()`/filter values serialize correctly (None → null, bool → true/false, dict/list → compact JSON), with 14 parametrized tests, all green. Happy to rebase or adjust if needed — could you take a look when you have a moment? Thanks!

PR #1610, opened 2026-09-03
fix(postgrest): retry transient failures for HEAD requests

### LINKED ISSUE #1609: PostgREST HEAD requests do not retry transient 503 or 520 responses
### Describe the bug

PostgREST `select(..., head=True)` requests do not retry transient Cloudflare 503 or 520 responses, even though HEAD is safe/idempotent and the retry helper documents support for both GET and HEAD.

`RequestConfig.should_retry()` checks for `"GET"` or `"HTTP"`; `"HTTP"` is not an HTTP method produced by the request builders, so HEAD always returns `False`.

### Reproduction

```python
from httpx import AsyncClient, Headers, QueryParams, Response
from yarl import URL

from postgrest.base_request_builder import RequestConfig

request = RequestConfig(
    AsyncClient(), URL("/example"), "HEAD", Headers(), QueryParams(), None, None
)
print(request.should_retry(Response(503), attempt_count=0))
```

This prints `False`. The equivalent GET request prints `True`.

### Steps to reproduce

1. Create a HEAD `RequestConfig` as above, or build one through `select(..., head=True)`.
2. Pass a 503 or 520 response to `should_retry()` before the retry limit.
3. Observe that the request is not eligible for retry.

### Library affected

postgrest

### Library version

postgrest 2.31.0 (current `main` at 5b8080a)

### Python version

Python 3.13.5

hsusul:
## Summary

- correct the retry method guard so HEAD requests are treated as safe/idempotent alongside GET
- cover both sync and async `select(..., head=True)` request builders

## Why

The retry helper documents GET and HEAD support, but the guard checked for the impossible method string `HTTP`. As a result, transient 503 and 520 responses to HEAD count requests failed immediately.

Closes #1609.

## Testing

- `uv run --package postgrest pytest src/postgrest/tests/_async/test_request_builder.py src/postgrest/tests/_sync/test_request_builder.py -q` (90 passed)
- `uv run --package postgrest ruff check src/postgrest/src/postgrest/base_request_builder.py src/postgrest/tests/_async/test_request_builder.py src/postgrest/tests/_sync/test_request_builder.py`
- `uv run --package postgrest ruff format --check src/postgrest/src/postgrest/base_request_builder.py src/postgrest/tests/_async/test_request_builder.py src/postgrest/tests/_sync/test_request_builder.py`
- `git diff --check`

### DIFF
--- a/src/postgrest/src/postgrest/base_request_builder.py
+++ b/src/postgrest/src/postgrest/base_request_builder.py
@@ -99,7 +99,7 @@ class RequestConfig(Generic[C]):
     def should_retry(self, response: RequestResponse, attempt_count: int) -> bool:
         if not self.retry_enabled or attempt_count >= MAX_RETRIES:
             return False
-        if not (self.http_method == "GET" or self.http_method == "HTTP"):
+        if not (self.http_method == "GET" or self.http_method == "HEAD"):
             return False
         return response.status_code == 503 or response.status_code == 520
 
--- a/src/postgrest/tests/_async/test_request_builder.py
+++ b/src/postgrest/tests/_async/test_request_builder.py
@@ -20,6 +20,15 @@ def test_constructor(request_builder):
     assert str(request_builder.path) == "/example_table"
 
 
+@pytest.mark.parametrize("head", [False, True])
+def test_select_request_retries_transient_response(
+    request_builder: AsyncRequestBuilder, head: bool
+):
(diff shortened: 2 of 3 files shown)


hsusul:
@olirice @silentworks — tagging you for review per the contribution guide. Thank you!

grdsdev (approved):

PR #1611, opened 2026-09-04
fix(postgrest): escape special chars in array filters so values with commas don't split

### LINKED ISSUE #1592: Array filters (cs/cd/contains/contained_by/overlaps) don't escape elements, corrupting values containing the comma delimiter
### Describe the bug

The array-membership filters `cs`, `cd`, `contains`, `contained_by`, and `overlaps` build their PostgreSQL array literal `{...}` from the raw `str()` of each element with no escaping:

https://github.com/supabase/supabase-py/blob/main/src/postgrest/src/postgrest/base_request_builder.py#L457

So any element that contains the comma delimiter (or a brace, double quote, backslash, whitespace, or that is empty or the word `NULL`) is emitted bare and mis-parsed by PostgREST. A value containing a comma is the clearest case: it silently splits into multiple array elements.

`in_()` already handles this correctly via `sanitize_param` (it quotes values containing `,:()`), but the `{...}` array operators do not.

### Reproduction

```python
qb = client.from_("things").select("*").contains("tags", ["a,b", "c"])
# qb sends: tags=cs.{a,b,c}
```

PostgREST parses `{a,b,c}` as **three** elements (`a`, `b`, `c`), not the two that were passed (`a,b` and `c`). The query silently matches the wrong thing instead of erroring.

The repo's own test suite documents the corruption: `test_contained_by_mixed_items` asserts that `contained_by("x", ["a", '["b", "c"]'])` produces `{a,["b", "c"]}`, which is itself a malformed array literal (its inner comma splits the second element apart).

### Expected behavior

Elements should be quoted following PostgreSQL array-literal rules (quote when empty, `NULL`, or containing a brace / comma / double quote / backslash / whitespace; backslash-escape embedded quotes and backslashes), so:

```
contains("tags", ["a,b", "c"])   ->   tags=cs.{"a,b",c}   # two elements, as passed
```

Simple values are unaffected (`["a", "b"]` stays `{a,b}`, `[1, 2, 3]` stays `{1,2,3}`).

### Library affected

postgrest

### Library version

main

### Python version

3.13

twelfthlabor:
#1592

`contains("tags", ["a,b", "c"])` was sending `tags=cs.{a,b,c}`, which Postgres parses as three elements (`a`, `b`, `c`) instead of the two I passed. No error, just silently wrong matches. Same issue in `cs`, `cd`, `contained_by`, and `ov`/`overlaps` since they all build the `{...}` literal the same way.

This PR adds a small `sanitize_array_element` helper in `utils.py` following the Postgres array literal rules - quote when empty, NULL, or containing braces, comma, quote, backslash, or whitespace, with backslash-escaping for quotes/backslashes inside. Simple values like `["a","b"]` and `[1,2,3]` come out exactly as before.

I verified it against a local Postgres:
- `'{"a,b",c}'::text[]` -> length 2, elem1 `a,b` (was 3 before)
- quote, backslash, empty string, NULL, space, and brace cases all round-trip correctly

One note: `test_contained_by_mixed_items` currently asserts the old malformed output `{a,["b", "c"]}`, which splits on the inner comma. That expectation will need updating to the quoted form.

Fixes #1592

### DIFF
--- a/src/postgrest/src/postgrest/base_request_builder.py
+++ b/src/postgrest/src/postgrest/base_request_builder.py
@@ -40,7 +40,7 @@ except ImportError:
 
 from .base_client import BasePostgrestClient
 from .types import JSON, CountMethod, Filters, JSONAdapter, RequestMethod, ReturnMethod
-from .utils import sanitize_param
+from .utils import sanitize_array_element, sanitize_param
 
 
 class QueryArgs(NamedTuple):
@@ -455,11 +455,11 @@ class BaseFilterRequestBuilder(Generic[C]):
         return self.filter(column, Filters.IN, f"({values})")
 
     def cs(self: Self, column: str, values: Iterable[Any]) -> Self:
-        values = ",".join(str(v) for v in values)
+        values = ",".join(sanitize_array_element(v) for v in values)
         return self.filter(column, Filters.CS, f"{{{values}}}")
 
     def cd(self: Self, column: str, values: Iterable[Any]) -> Self:
-        values = ",".join(str(v) for v in values)
+        values = ",".join(sanitize_array_element(v) for v in values)
         return self.filter(column, Filters.CD, f"{{{values}}}")
 
     def contains(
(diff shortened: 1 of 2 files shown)


twelfthlabor:
Closing in favor of #1593, which covers the same array-literal escaping with tests in both async and sync mirrors plus ruff/mypy. Happy to help review or test that one.

PR #1658, opened 2026-09-22
fix(storage): normalize content-type option casing

### LINKED ISSUE #37315: Storage API: Case-sensitive "Content-Type" header causes multipart boundary storage instead of file content
Supabase storage API stores multipart form data boundaries instead of actual file content when using capitalized "Content-Type" header. Files upload successfully (200 OK) but contain the entire multipart envelope rather than just the file bytes.

Upload file using capitalized header:

```python
# ❌ This causes the issue
headers = {"Content-Type": file.content_type, "upsert": "true"}
```

File gets stored with multipart boundaries:
```
--boundary123\r\nContent-Disposition: form-data; name="file"; filename="test.pdf"\r\nContent-Type: application/pdf\r\n\r\n%PDF-1.4...
```

Downloads return malformed content instead of actual file

Expected behavior
File should be stored with just the file content, regardless of header capitalization.
Actual behavior

Capitalized "Content-Type": Stores multipart boundaries ❌
Lowercase "content-type": Stores file content correctly ✅

**Environment**

Supabase Python client: v2.17.0
Environment: Local development
Python version: 3.12.11

**Additional context**

- HTTP RFC states headers should be case-insensitive
- This affects both Python client and direct REST API calls
- Related to multipart form parsing in storage backend
- No error thrown - files appear to upload successfully but contain wrong data

**Minimal Replication**

```
from fastapi import UploadFile
from supabase import create_client

# This breaks
result = supabase.storage.from_("bucket").upload(
    filename, 
    file_content,
    file_options={"Content-Type": file.content_type}  # ❌
)

# This works  
result = supabase.storage.from_("bucket").upload(
    filename,
    file_content, 
    file_options={"content-type": file.content_type}  # ✅
)
```

kekekuli:
## Why this needed?
  See my comments at [supabase/supabase#37315.](https://github.com/supabase/supabase/issues/37315#issuecomment-5773427511)
  
  When a capitalized `Content-Type` is passed in `file_options`, it is not extracted as the multipart file
  content type. Instead, it remains in the request-level headers and overrides HTTPX's generated
  `multipart/form-data` header.

  This causes the Storage service to store the complete multipart body, including its boundaries, as the
  file content.

  Fixed supabase/supabase#37315.

  ## What is the new behavior?

  The `content-type` file option is now matched case-insensitively before being merged with the default
  options.

  Only `content-type` is normalized. Other option keys retain their original casing.

  This applies to:

  - Synchronous and asynchronous clients

  Added regression tests also.

  ## Additional context

  The generated requests were captured before and after the fix.

  Before:

  ```text
  Request Content-Type: application/pdf
  File part Content-Type: text/plain;charset=UTF-8

  After:

  Request Content-Type: multipart/form-data; boundary=...
  File part Content-Type: application/pdf

  Test results:

  - 32 Storage client and utility tests passed
  - Ruff lint and format checks passed
```

  ### AI usage

  OpenAI Codex was used to investigate the HTTPX multipart request construction, assist with the
  implementation and regression tests, I did the review and some edits for outputs. The behavior was also verified by
  directly capturing and comparing the generated requests by me.

### CODE REVIEW

File: src/storage/src/storage3/utils.py [hunk 1]
```diff
@@ -1,10 +1,23 @@
...
+from .constants import DEFAULT_FILE_OPTIONS
 from .version import __version__
 
 
+def merge_file_options(
+    file_options: Optional[Mapping[str, Any]],
+) -> dict[str, Any]:
+    options: dict[str, Any] = {**DEFAULT_FILE_OPTIONS}
+    for key, value in (file_options or {}).items():
+        normalized_key = "content-type" if key.lower() == "content-type" else key
```
kekekuli: I am wondering if needed to use key.lower() for all keys, I am not a big fan to special case. Is it will break things? Wating your feedbacks.

src/storage/src/storage3/utils.py
```diff
@@ -1,10 +1,23 @@
...
 
+from .constants import DEFAULT_FILE_OPTIONS
 from .version import __version__
 
 
+def merge_file_options(
+    file_options: Optional[Mapping[str, Any]],
+) -> dict[str, Any]:
+    options: dict[str, Any] = {**DEFAULT_FILE_OPTIONS}
+    for key, value in (file_options or {}).items():
```
kekekuli: The original issue happened because top-level `file_options` were not validated, so unrecognized keys leaked into request headers. This fix special-cases `Content-Type`, but similar cases such as `{"Cache-
  Control": "7200"}` or `{"Upsert": "true"}` remain.

 Could we follow `storage-js` by accepting only known top-level options and requiring custom request headers under `headers`? We should decide whether unknown options should be rejected or ignored for backward compatibility.

kekekuli:
@grdsdev Hello, could you take a look? :)

